Issue #9, opened 2016-10-20
read_shared_strings is broken where there's RichTextRun cells


brunoqc:
I have a big (14M) excel file with about 20 sheets. I only need to read one sheet and I can validate everything but on the last line some of the strings doesn't match. I have the same row count.

I'll check my Go implementation if I have the same problem and if I did something special.

brunoqc:
I'm working on this.

brunoqc:
If I understand [Working with the shared string table (Open XML SDK)](https://msdn.microsoft.com/en-us/library/office/gg278314.aspx) correctly, if you have more than one style in a cell, you have multiples `<t>`.

Before my commit, they were all added to the vector so the value added after weren't accessible with the right index.

PR #10, opened 2016-10-20
Support RichTextRun shared strings

brunoqc:
Note that I discard the style (font...). Maybe you want a DataType::RichText.


brunoqc:
That's weird, I thought the PR would be mentioned in #9 automatically.

tafia:
Awesome!
I don't necessarily want a new type, what you've done looks good to me.

One point though, I'd prefer having a `rich_buffer: Option<String>` to replace both the `String` and the flag. This feel more rustic to me and we pay for the `String` allocation only when needed.

brunoqc:
I pushed a new commit using Option<String>.

> I'd prefer having a rich_buffer: Option<String> to replace both the String and the flag. This feel more rustic to me and we pay for the String allocation only when needed.

Yes it's better and I doesn't need to clone the string when I add it to the vector.

I also changed the unit test to ensure that rich_buffer is reset if there's more than one <si> with rich text.

tafia:
Thanks for your time!

PR #11, opened 2016-10-20
add DataType::Error

brunoqc:
I don't know if every errors are represented as a string.


tafia:
Thanks!
I am wondering if we shouldn't have an enum instead of a String for the excel errors? Given that we know all kind of possible errors.
What do you think?

brunoqc:
I agree that an enum would be better.

I updated the PR.

I found a [list of possible errors](https://msdn.microsoft.com/en-us/library/office/ff839168.aspx) but  no explanation so I translated the error messages from Excel.

I used the `unimplemented!()` macro in case we get an error we don't know, maybe we should use `unreachable!()` if we think we know *every* errors type. Maybe we shouldn't use either if they aren't in rust stable yet (I didn't test since I think office doesn't run with rust stable).

I wrote a function to parse the error string. It may be better since we can write a test for it like I did. Not sure if it's better to test a xlsx file with the errors or just the function or both. Maybe the function could be inline too.

Not sure if I should name the things Error or CellError. I don't want to confuse people with errors.rs.

summary

- is it OK to use a function (parse_error)
- test only a xlsx file, only the function or both
- is unimplemented ok
- CellError(CellErrorType) or Error(ErrorType)

tafia:
There is a conflict because of #10 merge. Could you rebase? (I you don't have time I'll do it, no worries).

All your decisions look good to me, this is what I had in mind!

brunoqc:
I'll rebase quickly.

brunoqc:
Done.

tafia:
:+1:

PR #24, opened 2016-10-25
Clean logic depending on file type

tafia:
The logic is internally split between several modules depending on the file type (xls, xlsx and xlsb).
This is much cleaner and prepares the ground for supporting xlsb and xls files.


tafia:
cc @brunoqc 
This cleans stuff quite a lot and adds a first support for reading xlsb.

PR #31, opened 2016-10-28
force exact dimensions on xlsx ranges

### LINKED ISSUE #29: bug when some rows has empty cells at the end
I have empty cells on some rows and it causes problems since office considers every row being the same length.

I wanted to fix it but I'm not sure how you would prefer it. Office could have a notion of rows in Range.inner. I'm not sure if we could add empty DataTypes in Range.inner since we would need to know the longest row's length and we would need to make 2 pass on the XML.

I can work on it if you lack the time.

I'll link a test case in which I also test for empty lines. I can split them in 2 test cases and I can also add them to issues.xlsx if you prefer.

```xml
<sheetData>
    <row r="1" spans="1:4" s="4" customFormat="1">
        <c r="A1" s="1" t="s">
            <v>0</v>
        </c>
        <c r="B1" s="2" t="s">
            <v>5</v>
        </c>
        <c r="C1" s="3" t="s">
            <v>10</v>
        </c>
        <c r="D1" s="4" t="s">
            <v>15</v>
        </c>
    </row>
    <row r="2" spans="1:4" s="4" customFormat="1">
        <c r="A2" s="1" t="s">
            <v>1</v>
        </c>
        <c r="B2" s="2" t="s">
            <v>6</v>
        </c>
        <c r="C2" s="3" t="s">
            <v>11</v>
        </c>
        <c r="D2" s="4" t="s">
            <v>16</v>
        </c>
    </row>
    <row r="3" spans="1:4" s="4" customFormat="1">
        <c r="A3" s="1" t="s">
            <v>2</v>
        </c>
        <c r="B3" s="2" t="s">
            <v>7</v>
        </c>
        <c r="C3" s="3" t="s">
            <v>12</v>
        </c>
    </row>
    <row r="4" spans="1:4" s="4" customFormat="1">
        <c r="A4" s="1" t="s">
            <v>3</v>
        </c>
        <c r="B4" s="2" t="s">
            <v>8</v>
        </c>
        <c r="C4" s="3" t="s">
            <v>13</v>
        </c>
        <c r="D4" s="4" t="s">
            <v>17</v>
        </c>
    </row>
    <row r="5" spans="1:4" s="4" customFormat="1">
        <c r="A5" s="1" t="s">
            <v>4</v>
        </c>
        <c r="B5" s="2" t="s">
            <v>9</v>
        </c>
        <c r="C5" s="3" t="s">
            <v>14</v>
        </c>
    </row>
</sheetData>
```

tafia:
closes #29 

I forced the size of the range to the dimension read in `/worksheet/dimension` and gets the `c[@r]` attribute to know the index within the range.

I consequently modified the tests accordingly (many empty rows as per the range dimension).

There may be an impact on performance (bigger output + parsing every cell reference) but I guess it is minimal.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -213,11 +213,11 @@ pub trait ExcelReader: Sized {
 impl Range {
 
     /// creates a new range
-    pub fn new(position: (u32, u32), size: (usize, usize), data: Vec<DataType>) -> Range {
+    pub fn new(position: (u32, u32), size: (usize, usize)) -> Range {
         Range {
             position: position,
             size: size,
-            inner: data,
+            inner: vec![DataType::Empty; size.0 * size.1],
         }
     }
 
@@ -229,6 +229,13 @@ impl Range {
     /// get size
     pub fn get_size(&self) -> (usize, usize) {
         self.size
+    }
+
+    /// set inner value
+    pub fn set_value(&mut self, pos: (u32, u32), value: DataType) {
+        assert!(self.position <= pos);
+        let idx = (pos.0 - self.position.0) * self.size.1 as u32 + pos.1 - self.position.1;
+        self.inner[idx as usize] = value;
     }
 
     /// get cell value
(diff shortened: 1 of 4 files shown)


tafia:
cc @brunoqc

brunoqc:
Seems fine :+1:

PR #32, opened 2016-10-28
test cases for special characters

brunoqc:
I based this commit on top of 1245ec4 in case you wanted to merge it.

The only problem that I have is for `< > and &`. The others case in the test are already OK but I thought it could be nice to have them. Maybe they could be spitted (the &amp; and other escaped characters vs the others).

I guess the changes would need to be made in quick-xml but it could be nice to have tests here too.


tafia:
Thanks
I won't merge failing test but I'll use it to implement special characters support ( via unescape methods on quick-csv).

PR #35, opened 2016-10-31
allow a xlsb sheet to have only empty rows

tafia:


tafia:
Also
- remove byteorder dependency for xlsb
- allow not having sharedStrings.bin file (if no cell saved)

Issue #20, opened 2016-10-24
Improve readme


tafia:
Adds several missing parts
- Examples
  - ~~Example for reading cells~~
  - ~~Example for reading vba, and checking references~~
- Better description of the library
- Benchmarks?
- Current limitations
  - ~~cannot read xlsb cells yet~~
  - cannot read xls cells yet
  - inconsistent vba string parsing (missing a proper MBSC encoding)
  - plus many missing functions
  - no writing, only read

PR #37, opened 2016-11-01
Better readme

tafia:
Better description / examples / disclaimers

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ A Excel file reader, in Rust.
 
 As long as your files are *simple enough*, this library is ready for use.
 
-## Usage
+## Examples
 
-### Simple example
+### Simple
 ```rust
 let mut excel = Excel::open("file.xlsx").unwrap();
 let r = excel.worksheet_range("Sheet1").unwrap();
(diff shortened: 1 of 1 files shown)


tafia:
cc @brunoqc would you have time to comment on this? 
English is not my first language ...

[rendered](https://github.com/tafia/office/blob/adc64387154f870134047ca3ff7e87e819a37322/README.md)

tafia:
Closes #20

brunoqc:
English is not my first language either and you write pretty well so it's very possible that your English is better than mine :). The README seems fine.

I like the examples in the README but they could be a bit intimidating to functional programming noobs (like me) I wonder if it would be a good idea also add a really simple exemple like:
```rust
let mut excel = Excel::open("file.xlsx").unwrap();
let r = excel.worksheet_range("Sheet1").unwrap();
for row in r.rows() {
    println!("row={:?}, row[0]={:?}", row, row[0]);
}
```

It could be a nice copy/paste starting point.

tafia:
Thanks!
Merging now

PR #39, opened 2016-11-02
Xls

### LINKED ISSUE #28: Read xls (CFB) cells
This is probably the only missing part before covering all excel files.

tafia:
Closes #28

This is a work in progress.
For now it splits all Compound File Binary format processing into its own module. Later i'll implement xls reading which should not be too complicated.


tafia:
@brunoqc fyi, office now reads all excel files!
Even if there is little test so lot of files will probably break.

brunoqc:
Cool! Good work!

Issue #27, opened 2016-10-27
Provide a lazy iterator over the rows


tafia:
Today the entire range is first saved in memory.
Depending on the need we could provide for both xlsx and xlsm a `Row` or even a `Cell` lazy iterator ... and collect them into a Range if needed.

brunoqc:
Do you think it would make sense to also have a non-reference iterator? I mean an iterator that would allow us to take ownership of the DataType without having to clone it. Maybe something like [Vec::drain](https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.drain).

There was also [Idiomatic way to take ownership of all items in a Vec<String>?](https://users.rust-lang.org/t/idiomatic-way-to-take-ownership-of-all-items-in-a-vec-string/7811) on the forum.

tafia:
> I realized before going to bed that it wouldn't make sense with the xml shared string concept.

Well, for this we could probably return a `Cow` with only strings being `Cow::Borrowed` ... even if it probably makes little sense to borrow the other primitives (pointer size anyway).

hwchen:
Just wanted to say that a lazy iterator would have been super useful for my use case.

I just needed to compare the headers (first row) for 13 sheets, each with (1,000,000 x 13) cells. I ended up having to load each complete range just to get the first row.

Fortunately, Rust and this library are pretty fast, it took about 2 minutes for what appears to be a 1G file.

Let me know if I missed something, perhaps there was a way for me to get the info I needed without loading everything.

tafia:
The implementation might depend on the actual file type (xlsx, xslb, ods etc ...) which one was yours just for info?

hwchen:
Mine was `xlsx`.

PR #61, opened 2016-11-16
Build range based on sparse Vec<Cell>

tafia:
Lot of sheets have formatting information lying on unused cells.
As we only want to retrieve values, we can skip those entirely.

Promoting the use of `Range::from_sparse` will be more efficient
- because of the smaller size of inner vector
- because of less checks on bounds
- because it avoids reallocating Range on sheets were dimension mismatch
with actual cell positions (happens mostly on old xls files)

Also 
- adds a `Range::unused` fn to iterate over unused cells
(directly on inner `Vec` instead off rows chunks).
- adds a `range_eq!` macro in tests to simplify test writing/reading
- fix range::get_value

This is a breaking change.

### DIFF
--- a/examples/search_errors.rs
+++ b/examples/search_errors.rs
@@ -2,6 +2,9 @@ extern crate calamine;
 extern crate glob;
 
 use std::env;
+use std::io::{BufWriter, Write};
+use std::fs::File;
+
 use glob::{glob, GlobError};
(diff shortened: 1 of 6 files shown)


tafia:
Also will ease #27 by postponing range creation at the very end.

PR #67, opened 2017-02-07
inlineStr implementation

pfernie:
This PR corrects behavior for xlsx parsing of inline string ("inlineStr") elements, which are encoded as ```<is>``` nodes per [this description](http://officeopenxml.com/SScontentOverview.php), instead of ```<v>``` nodes. Also, in some sheets I work with sheetname references already contain a "xl/" or "/xl/" prefix; a patch to handle this is also contained.

The PR also includes some unrelated commits:
* rustfmt formatting updates
* bump error-chain dependency to 0.8.1 (no code changes needed)
* added some debug!() statements

### DIFF
--- a/src/xlsx.rs
+++ b/src/xlsx.rs
@@ -5,7 +5,7 @@ use std::borrow::Cow;
 
 use zip::read::{ZipFile, ZipArchive};
 use zip::result::ZipError;
-use quick_xml::{XmlReader, Event, AsStr};
+use quick_xml::{XmlReader, Event, AsStr, Element};
 
 use {DataType, ExcelReader, Range, Cell};
 use vba::VbaProject;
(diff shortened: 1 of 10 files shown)


src/xlsx.rs
```diff
@@ -100,7 +100,19 @@ impl ExcelReader for Xlsx {
...
+                                // target may have pre-prended "/xl/" or "xl/" path;
+                                // strip if present
+                                let r = if r.starts_with("/xl/") {
+                                    &r[4..]
+                                } else if r.starts_with("xl/") {
+                                    &r[3..]
+                                } else {
+                                    r
+                                };
+                                path = format!("xl/{}", r);
```
tafia: Not a big dead but we could avoid some `format!` here
```rust
                                path = if r.starts_with("/xl/") {
                                    r[1..].to_string()
                                } else if r.starts_with("xl/") {
                                    r.to_string()
                                } else {
                                    format!("xl/{}", r)
                                };
```

src/xlsx.rs
```diff
@@ -190,58 +190,113 @@ fn read_sheet_data(xml: &mut XmlReader<BufReader<ZipFile>>,
...
-                        Ok((b"r", v)) => Some(get_row_column(v)),
-                        _ => None,
-                    })
-                    .next() {
-                    Some(v) => v?,
-                    None => return Err("Cell without a 'r' reference tag".into()),
-                };
+                let pos = get_attribute(c_element, b"r")
+                    .and_then(|o| o.ok_or_else(|| "Cell missing 'r' attribute tag".into()))
+                    .and_then(get_row_column)?;
```
tafia: Much easier to read indeed

src/xlsx.rs
```diff
@@ -190,58 +190,113 @@ fn read_sheet_data(xml: &mut XmlReader<BufReader<ZipFile>>,
...
-                                        _ => v.parse().map(DataType::Float)?,
-                                    };
-                                    cells.push(Cell::new(pos, value));
+                                    cells.push(Cell::new(pos, read_value(xml, strings, c_element)?));
                                     break;
                                 }
-                                b"f" => (), // formula, ignore
-                                _name => return Err("not v or f node".into()),
+                                b"f" => {} // ignore f nodes
+                                n => return Err(format!("not a v, f, or is node: {:?}", n).into()),
```
tafia: if possible can you add some quote (because of the *not a*, I find it a little more complicated to read:
`not a 'v, 'f' or 'is' node`

src/xlsx.rs
```diff
@@ -172,61 +186,134 @@ impl ExcelReader for Xlsx {
...
+                                b"is" => {
+                                    cells.push(Cell::new(pos, read_inline_str(xml)?));
+                                    break;
+                                }
+                                b"v" => {
+                                    cells.push(Cell::new(pos, read_value(xml, strings, c_element)?));
+                                    break;
+                                }
+                                b"f" => {} // ignore f nodes
+                                n => return Err(format!("not a v, f, or is node: {:?}", n).into()),
```
tafia: if possible can you add some quote (because of the not a, I find it a little more complicated to read:
not a 'v, 'f' or 'is' node

tafia:
Published version 0.5.0.
Thanks again!

PR #70, opened 2017-03-06
Update dependencies and move from rust-encoding to encoding_rs

tafia:
Deps to fix some bugs
Encoding_rs to speed decoders (see https://github.com/tafia/bench_decode)


tafia:
This is a breaking change as some decoders are not supported anymore. I believe they are not widely used anyway.

PR #71, opened 2017-03-11
[RFC] let sheet_names return vector of String instead of &str

### LINKED ISSUE #69: lifetime issue when itering over the `sheet_names()`
Here is the code:
``` rust
    let mut xl = Excel::open(&sce).unwrap();
    for stn in xl.sheet_names().unwrap() {
        let range = xl.worksheet_range(stn).unwrap();
        write_range(&mut dest, range).unwrap();
    }
```
I want to iter over the sheets with sheet names. But when I compile it, I got lifetime issue:

```
error[E0499]: cannot borrow `xl` as mutable more than once at a time
  --> src/main.rs:28:21
   |
27 |     for stn in xl.sheet_names().unwrap() {
   |                -- first mutable borrow occurs here
28 |         let range = xl.worksheet_range(stn).unwrap();
   |                     ^^ second mutable borrow occurs here
29 |         write_range(&mut dest, range).unwrap();
30 |     }
   |     - first borrow ends here

error: aborting due to previous error
```

Is there any way to fix it easily?

grissiom:
*Note that this is a API-breaking change*

&str will have the same lifetime with &self. This will make it hard to
use in iterations like this:

```
let mut xl = Excel::open(&sce).unwrap();
for stn in xl.sheet_names().unwrap() {
	let range = xl.worksheet_range(stn).unwrap();
	write_range(&mut dest,  range).unwrap();
}
```

Making it return vector of String could fix this.

This is a attempt to fix #69 . However, I'm new to Rust so please review it hardly ;)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -273,9 +273,9 @@ impl Excel {
     /// let mut workbook = Excel::open(path).unwrap();
     /// println!("Sheets: {:#?}", workbook.sheet_names());
     /// ```
-    pub fn sheet_names(&mut self) -> Result<Vec<&str>> {
+    pub fn sheet_names(&mut self) -> Result<Vec<String>> {
         self.initialize()?;
-        Ok(self.sheets.iter().map(|&(ref k, _)| &**k).collect())
+        Ok(self.sheets.iter().map(|&(ref k, _)| k.to_string()).collect())
     }
 }
 


src/lib.rs
```diff
@@ -273,9 +273,9 @@ impl Excel {
     /// let mut workbook = Excel::open(path).unwrap();
     /// println!("Sheets: {:#?}", workbook.sheet_names());
     /// ```
-    pub fn sheet_names(&mut self) -> Result<Vec<&str>> {
+    pub fn sheet_names(&mut self) -> Result<Vec<String>> {
         self.initialize()?;
-        Ok(self.sheets.iter().map(|&(ref k, _)| &**k).collect())
+        Ok(self.sheets.iter().map(|&(ref k, _)| (**k).to_string()).collect())
```
tafia: You probably don't need to manually dereference `**k`, it should be done automatically.

tafia:
@grissiom do you have any comment to do on my answer?

grissiom:
Sorry for the delay. I just amend the commit to fix the dereference issue.

I think return `String` is OK here since there won't be too many sheets and the sheet names should be considered small compared to the data inside it. So it should be cheap.

I think make `sheet_names` more ergonomic is better ;)

tafia:
Thanks

grissiom:
Thank you for your patients and the effort to make Excel available for Rust!

tafia:
You're welcomed. Happy that it works:)

PR #72, opened 2017-03-15
Support richtext in inlineStr

pfernie:
The original support for inlineStr does not handle richtext. This patch factors out the richtext logic in `read_shared_string`, and re-uses it for parsing inlineStr cells.

### DIFF
--- a/src/xlsx.rs
+++ b/src/xlsx.rs
@@ -57,8 +57,17 @@ impl ExcelReader for Xlsx {
             Some(x) => x?,
         };
         let mut strings = Vec::new();
-        while let Some(s) = read_string(&mut xml, b"si")? {
-            strings.push(s);
+        let mut buf = Vec::new();
+        loop {
+            match xml.read_event(&mut buf) {
+                Ok(Event::Start(ref e)) if e.name() == b"si" => {
+                    if let Some(s) = read_string(&mut xml, b"si")? {
+                        strings.push(s);
+                    }
+                }
+                Ok(Event::End(ref e)) if e.name() == b"sst" => break,
+                _ => (),
+            }
         }
         Ok(strings)
     }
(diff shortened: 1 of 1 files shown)


src/xlsx.rs
```diff
@@ -409,6 +370,37 @@ fn get_row_column(range: &[u8]) -> Result<(u32, u32)> {
...
+            }
+            Ok(Event::Start(ref e)) if e.name() == b"t" => {
+                let value = xml.read_text(b"t", &mut Vec::new())?;
+                if let Some(ref mut s) = rich_buffer {
+                    s.push_str(&value);
+                } else {
+                    return Ok(Some(value));
+                }
+            }
+            Ok(Event::Eof) => return Ok(rich_buffer),
```
tafia: shouldn't this be an error (we haven't found the closing event)

pfernie: This & the test failures were related; I apparently did not re-run the tests after merging the shared string & inlineStr cases, and missed a logic error that introduced. This case should indeed be an error; corrected in lastest patch.

src/xlsx.rs
```diff
@@ -60,34 +58,16 @@ impl ExcelReader for Xlsx {
...
-                    } else {
-                        strings.push(value);
-                    }
-                }
-                Ok(Event::Eof) => break,
-                Err(e) => return Err(e.into()),
+                Ok(Event::End(ref e)) if e.name() == b"sst" => break,
                 _ => (),
             }
-            buf.clear();
```
tafia: you should probably add `clear` back to keep memory low.

src/xlsx.rs
```diff
@@ -409,6 +379,37 @@ fn get_row_column(range: &[u8]) -> Result<(u32, u32)> {
...
+            }
+            Ok(Event::End(ref e)) if e.name() == closing => {
+                return Ok(rich_buffer);
+            }
+            Ok(Event::Start(ref e)) if e.name() == b"t" => {
+                let value = xml.read_text(b"t", &mut Vec::new())?;
+                if let Some(ref mut s) = rich_buffer {
+                    s.push_str(&value);
+                } else {
+                    return Ok(Some(value));
```
tafia: Perhaps we should call [`xml.read_to_end(closing, &mut Vec::new())?`](https://docs.rs/quick-xml/0.6.2/quick_xml/reader/struct.Reader.html#method.read_to_end) first ?

pfernie:
Latest commit addresses your 2 comments; tests are passing and working for my use case. However, in putting together a new sample .xlsx for a test case, I ran into issues for which I will open a new PR.

tafia:
Thanks!

PR #84, opened 2017-04-24
add EUC_KR encoding

tafia:
Following https://github.com/tafia/calamine/commit/09de5859fc78e981c6570f4fcc46f5d4f0017c6d#commitcomment-21867087 comment

@hsivonen do you have a file to test it?


hsivonen:
Sorry, I don't have a windows-949-encoded Excel file. I was just taking a look at how other crates use encoding_rs and noticed this.

(I haven't checked if EUC-KR in the Encoding Standard matches windows-949 exactly, but in any case it's very close. It includes the original KS X 1001 part that the original EUC-KR used and the Unified Hangul Code part that windows-949 added.)

tafia:
Ok. I am not very familiar with encodings. I'll merge as it then (probably better than not supporting it at all anyway).
Thanks again!

PR #94, opened 2017-07-07
Add an additional bound check inside cfb::decompress_stream

pgkos:
This adds an additional bound check to `cfb::decompress_stream`. Previously, there was a panic in the line `let bit_flags = s[i]`.


tafia:
Thanks!

tafia:
Published v0.11.7

Issue #96, opened 2017-07-26
Serde serialization and deserialization?


erickt:
Hello!

Have you considered adding support for [serde](https://serde.rs) support for serialization and deserialization, at least for the `DataType` type? It seems like it would be pretty straightforward. Are there any downsides to doing this that I'm not thinking of?

tafia:
I'd love adding serde support. 

Unfortunately it doesn't look that trivial to me, for instance, I am deserializing several different file format into `DataType`, wouldn't this be an issue?

I haven't looked at it yet to be honest.

tafia:
Closing as the serialization part needs calamine to be able to write an excel file.
Thanks again @erickt

PR #99, opened 2017-10-17
Implement a serde deserializer

erickt:
This draws inspiration from @burntsushi's [csv](https://github.com/burntsushi/rust-csv). This adds a `RangeDeserializerBuilder`, `RangeDeserializer`, and `ToCellDeserializer`, which work together to iterate over each row in a `Range`, all while deserializing the row into some deserializable type.

Some questions:

* One thing to note is that it assumes that the first row is the header. One needs to use the Builder to disable this. Is this the right choice?
* Should the builder be exposed on `Range`?
* It feels like I ought to be working with `Cell`s, not `CellType`s, but it doesn't seem like cells are getting much use.
* Should `DataTypeDeserializer` be exposed more directly, where it is constructed with a (0,0) position?

Implements the deserialization part of #96.

Also, this bumps the error-chain dependency to 0.11.


tafia:
This is awesome thanks!! 
I do not have lot of time this week but I plan to review it properly next week.
Thanks!

erickt:
Happy to help! calamine was a great help processing the rust community survey, so it was the least I could do :)

tafia:
Very happy to hear that!

tafia:
This is very, very, very nice thanks again!

I must confess I do not fully understand it all. It looks like there are a lot of intermediary struct/trait I wouldn't have used. 

That said: 
- I am really no serde expert, 
- it is not breaking anything and all doc tests pass
- it adds a very wanted feature to calamine
- you and @BurntSushi are probably far better programmer than me

I'm happy to merge, I'll release a new version asap.
Thanks again, this is no small work.

tafia:
version 0.12.0 published!

PR #104, opened 2017-12-31
document the coordinates in `get_size` and `get_value`

grissiom:
The end user may hard to see which is row and which is column from (u32,
u32). So document them in the doc string.


tafia:
Thanks!
I won't publish a new version just for that though but the docs will be updated next time I'll do it, which may not be that long.

grissiom:
Yes, just a small doc fix like this doesn't deserve a new version. No hurry ;-)

Issue #100, opened 2017-11-20
Feature/API request: Open sheets from memory buffers


golddranks:
At the moment, the `open` API seems to be the only way to open a new workbook. However, it would be good to have also a way to open sheets directly from a memory buffer, without accessing the hard disk.

Imagine that an user has uploaded a smallish XLSX file to your web service, and the backend is written with Rust. You don't want to first save the file to disk (an operation that might have a worst case latency of tens of milliseconds), and then load it there to open it up – instead, you want to directly open the file from a `&[u8]`.

tafia:
I agree it would be more convenient.

It is not that simple though:
- I need to know the extension => we can probably have one method per file
- for xls[xmb], it needs to unzip the file. I am using `zip` crate which [requires a `Read + Seek`](https://docs.rs/zip/0.2.6/zip/read/struct.ZipArchive.html)
- I need to know before hand the size length

All of this to say that it would be probably a good idea to have a trait that is providing all this and implement that trait for `File` plus some custom struct built upon `&[u8]`.

sanmai-NL:
@tafia: I also have this requirement. I’ll address the obstacles you mentioned:

- Expose an enum with the appropriate file formats, e.g. the existing `FileType` or some alternative, as require it as parameter to a `from_reader()` method.

- That’s not a problem with [`std::io::Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html)
```rust
let reader = Cursor::new(bytes as &[u8]);
zip::ZipArchive::new(reader)
```

- Size length? The length of a slice is the [`len()`](https://doc.rust-lang.org/std/primitive.slice.html#method.len) method. Did you mean something more complex?

sanmai-NL:
I implemented it, will open a PR soon. We can then discuss it there.

tafia:
Feel free to reopen it if needed!

ddimaria:
For anyone searching for loading an excel file from memory (a vec in this case):
```rust
let cursor = Cursor::new(file);
let mut reader: Xlsx<_> = ExcelReader::new(cursor).unwrap();
if let Ok(r) = reader.worksheet_range("Sheet1") {
    for row in r.rows() {
        println!("row={:?}, row[0]={:?}", row, row[0]);
    }
}
```

PR #105, opened 2018-01-04
Process readers, not only files

sanmai-NL:
This is more generic and useful in a networking context. It also helps us to with fuzz testing the parsers. See #100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,11 @@ It is as simple as:
 
 ```rust
 use calamine::{Result, Sheets, RangeDeserializerBuilder};
+use std::io::File;
 
 fn example() -> Result<()> {
     let path = format!("{}/tests/tempurature.xlsx", env!("CARGO_MANIFEST_DIR"));
-    let mut workbook = Sheets::open(path)?;
+    let mut workbook = Sheets::<File>::open(path)?;
     let range = workbook.worksheet_range("Sheet1")?;
 
     let mut iter = RangeDeserializerBuilder::new().from_range(&range)?;
@@ -45,7 +46,9 @@ fn example() -> Result<()> {
 
 ### Reader: Simple
 ```rust
-let mut excel = Sheets::open("file.xlsx").unwrap();
+use calamine::Sheets;
+
+let mut excel = Sheets::<File>::open("file.xlsx").unwrap();
 let r = excel.worksheet_range("Sheet1").unwrap();
 for row in r.rows() {
     println!("row={:?}, row[0]={:?}", row, row[0]);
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -152,15 +154,15 @@ impl fmt::Display for CellErrorType {
 }
 
 /// File types
-enum FileType {
+enum FileType<RS> where RS: Read + Seek {
```
sanmai-NL: We can remove the `Seek` bound if `Xls` is implemented without seeking. If the API changes to split the `FileType` and the current tuple struct member of each variant, we can immediately remove the `Seek` bound for all formats other than `Xls`.

src/lib.rs
```diff
@@ -225,14 +228,31 @@ impl Sheets {
         })
     }
 
+    /// Creates a new workbook from a reader.
+    /// ```
+    pub fn new(reader: RS, extension: &str) -> Result<Sheets<RS>> where RS: Read + Seek {
```
sanmai-NL: I prefer to use an `enum` for the spreadsheet formats rather than filename extensions. We can map filename extensions to format variants in a separate function (if the `open` constructor is to remain).

README.md
```diff
@@ -24,10 +24,11 @@ It is as simple as:
 
 ```rust
 use calamine::{Result, Sheets, RangeDeserializerBuilder};
+use std::io::File;
 
 fn example() -> Result<()> {
     let path = format!("{}/tests/tempurature.xlsx", env!("CARGO_MANIFEST_DIR"));
-    let mut workbook = Sheets::open(path)?;
+    let mut workbook = Sheets::<File>::open(path)?;
```
sanmai-NL: Maybe the explicit `File` type parameter requirement can be avoided somehow.

tafia: It doesn't really bother me

src/lib.rs
tafia: Is this really important to remove it? We can use a `Cursor` for if needed.

src/lib.rs
tafia: I agree it would probably be better

src/lib.rs
sanmai-NL: Using `Cursor` is only possible if the full file is in memory. Maybe not critical, but this bound should be reviewed before stabilizing the API.

src/lib.rs
sanmai-NL: I have a solution tested in practice, that works will. We can accept a Internet Media Type (‘MIME type’) instead of an extension.

PR #107, opened 2018-01-10
Update log from 0.3.8 to 0.4

laumann:
In the minor version bump log::LogLevel has been renamed to log::Level, so this
is changed as well.


laumann:
Not sure what to do about the failing tests - they fail with and without this change.

tafia:
I think there is some regression on zip crate. Forcing it to `=0.2.6` fixes it.

tafia:
Thanks

laumann:
Thanks! I had the same problem with zip. I ended up editing my Cargo.lock to force it to build calamine with zip 0.2.6

PR #109, opened 2018-01-13
zip: Enable feature "deflate"

laumann:
This is to work around zip 0.2.9 turning "deflate" into a feature and
calamine disabling all features by default.

See also mvdnes/zip-rs#23


tafia:
Thanks!

laumann:
No problem :-) It's kind of unfortunate that the `deflate` feature was added in a patch release of `zip`, because it's not backwards compatible (users of <0.2.9 of zip will have to update to 0.2.9)

Issue #101, opened 2017-11-20
Don't clone on returning the sheet names


golddranks:
The `sheet_names` method clones the vector of the sheet names. I get that this isn't likely to be on the hot path, but cloning and allocating should usually left to the user when possible. (Nothing wrong with convenience methods though). As an alternative, it would be a good practice to provide an iterator over the sheet names: `sheet_names_iter`, with `Item=&str`. This avoids both allocation and cloning, leaving the maximum flexibility in the hands of the user.

tafia:
We could provide an additional method indeed. Do you want to make a PR?

golddranks:
I'll send a PR during the next week!

sanmai-NL:
@golddranks: Do you have time to pick this up again?

tafia:
I implemented it in #111

PR #111, opened 2018-01-23
Migrate to Failure crate

### LINKED ISSUE #110: Migrate to failure
Also type all string errors into enum variants.

tafia:
This is a big breaking change and comes with lot of cleaning.
Each document has its own error enum and CalError is wrapping them all.
Every `bail!` call have been removed for a dedicated enum variant.

Closes #110


tafia:
There is also a change on `Reader` trait which now allows statically using the right (say Xlsx) reader.

tafia:
This refactoring closes #101 as well.

Issue #114, opened 2018-02-04
Bug(s) in indexing worksheets


sanmai-NL:
#113 contains code aimed at debugging indexing bugs.

```rust
let range = workbook.worksheet_range(name_worksheet.as_str());
if range.is_empty() {
    continue
}
let (row_max, col_max) = (row_max_usize as u32, col_max_usize as u32);
trace!("{}x{} matrix", row_max, col_max);
for index_col in 0..col_max {
    for index_row in 0..row_max {
        // panics
        let cell = range.get_value((index_row, index_col));
    }
}
```

sanmai-NL:
# Empty worksheet considered 1x1
```
TRACE 2018-02-04T14:34:21Z: calamine_debug: 1x1 matrix
thread 'calamine_debug::tests::xls_1' panicked at 'assertion failed: self.inner.len() > 0', calamine/src/lib.rs:411:9
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:380
   3: std::panicking::default_hook
             at libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:576
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:537
   6: <calamine::Range<T>>::get_value
             at 
<panic macros>:3
   7: calamine_debug::n_personaldatum_from_reader
             at src/personaldatum/semistructured.rs:100
```

sanmai-NL:
## Column 0 not accessible in 30x6 worksheet

```
TRACE 2018-02-04T14:34:21Z: calamine_debug: 30x6 matrix
thread 'calamine_debug::tests::xls_3' panicked at '0 - 1 is out of bounds.', calamine/src/lib.rs:413:87
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
[...]

PR #119, opened 2018-03-27
Fix indexing

tafia:
Fix various bugs listed in #114 

- [x] Empty worksheet considered 1x1
- [x] Column 0 not accessible in 30x6 worksheet


tafia:
@sanmai-NL do you have a test file for the 30x6 failing test.
Can you also check if your issue is fixed for the empty sheet?

sanmai-NL:
@tafia: Sent you an e-mail with more info. :slightly_smiling_face:

tafia:
Thanks @sanmai-NL 

Sorry for the wait.

I have tested your files and I managed to read them all (at least with this PR).

I think there is an error with your code in #114, instead of using `Range::get_value`, you should be using the `Index` trait directly, i.e. `range[(0, 0)]`.

Indeed the `get_value` expects an **absolute** value whereas in your example the range value were starting at row 1 (not 0) only.

I have modified the `get_value` fn so it'll return an `Option` instead. I have also improved the comments so hopefully it'll be clearer. I couldn't find a better name though (which is not too long).

sanmai-NL:
Thanks! I had some issues using relative indexes and I also wanted to prevent panics. Thanks for improving the code, I will try it out soon.

tafia:
>  I had some issues using relative indexes

Just to be sure I was clear. If you want to continue using absolute indexes, you must loop from `start` to `end` instead of from `(0, 0)` to `size`.

Let me know!

sanmai-NL:
Ah okay, that clarifies things.

PR #124, opened 2018-07-24
Fix boolean detection and missing repeated cells in ODS

zetok:
ODS doesn't store repeated cells, but uses an attribute to inform
how many repeated cells there are.

As for the boolean detection, I only added lower-case boolean to look
for, since it is what is in my spreadsheet.

I don't really know whether the code is the way it is supposed to be,
but it does work for me and appears to fix bugs I found.


tafia:
Thanks a lot I totally forgot it, sorry!!!

tafia:
Published v0.14.2

Issue #127, opened 2018-08-21
Feature request: original order preservation for Ods::sheet_names()


kklibo:
Great library, it's saved me countless hours of work. I'm using it to read ods spreadsheets, and I could use a list of sheet names in the original order that they appeared in the file.

tafia:
Sorry for the late answer.
This is indeed a good idea. Not sure yet how it works though. I you have some time, feel free to give it a try yourself, I'd be happy to merge it. I could also help mentoring on calamine if needed.

kklibo:
Thanks. It looks like the order is determined by the hash table it's stored in, so it's basically random now. I'll take a look this weekend.

tafia:
We can have a `BTreeMap` if needed. The speed difference should be ridiculous.

masklinn:
@tafia other than historical is there any reason that would not be the default everywhere? It doesn’t seem to be the case in xls files (either), and working in “the first sheet” seems pretty common, either that being the only worksheet or other worksheets being ancillary or documentation and such.

tafia:
No there is no other reason than historical.

I didn't spend time on that particular point because I never needed it but I'd welcome any pull request.

masklinn:
> No there is no other reason than historical.
> 
> I didn't spend time on that particular point because I never needed it but I'd welcome any pull request.

Great, I’ll find some time to do that then. Would you mind a shortcut to get a sheet by index as well?

PR #128, opened 2018-08-26
record ods sheet names in order as they are loaded

kklibo:
(instead of reading them from the sheet hashmap)

### DIFF
--- a/src/ods.rs
+++ b/src/ods.rs
@@ -106,9 +106,9 @@ impl<RS: Read + Seek> Reader for Ods<RS> {
             Err(e) => return Err(OdsError::Zip(e)),
         }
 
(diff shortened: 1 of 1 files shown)


kklibo:
re: #127 
The sheet name list was already being saved at load time, so it was trivial to save it in order as it was loaded instead of reading it out of the HashMap.

The order is now determined by the loop in ods::parse_content, so only use this commit if that is guaranteed to load the sheets in their correct order. I assume it would, and it seems to work in a trivial test case, but if there's some aspect of the ods format that specifies a different order for worksheets, this fix doesn't account for it.

tafia:
Thanks for the PR!
Could you please rebase on master?
I have updated the travis script.
Can you also add a test?

PR #133, opened 2018-09-05
Parse cells which contains 'text:p' or 'text:s' tags

gzsombor:


gzsombor:
What do you think about this?

tafia:
Sorry for the late response and thank you for the PR!
Looks great to me, thanks!

tafia:
pushed v0.14.7

PR #140, opened 2018-12-02
Use codepage/encoding_rs for codepage mapping and single-byte check

hsivonen:
This should not change observable behavior. It just uses encoding_rs and its companion codepage crate for some operations that they provide.

The codepage crate knows about the 21010 special case.


hsivonen:
Oops. I suggest s/codpage/codepage/ when squashing.

tafia:
Thanks !!
I'm on mobile but I'll publish it asap!

PR #144, opened 2018-12-14
Make it more user friendly

### LINKED ISSUE #141: Documentation enhancement: How do I query a range of cells?
The examples I found all use the whole of sheet1 as the range. I have an Excel file that has a header row, followed by data rows grouped into sets of 4 columns each. 

For external reasons I can't use export to CSV, or change the spreadsheet in any way.

I'd like to be able to find an example in the documentation describing how to convert the string b"A2:D91" to a Range that can be used within the Serde Deserialization example, instead of getting the range via workbook.worksheet_range("Sheet1")

tafia:
- `DataType`: 
  - impl `PartialEq` for easy filtering
  - add convenient `is_...` and `get_...` function to avoid destructuring (one-liners)
- iterators
  - add `size_hint`s (perf)
  - add `DoubleEndedIterator` (perf when searching from the end)
  - add `ExactSizeIterator` useful when collecting
  - add a new simple `Cells` iterator ... which iterates over empty and non empty cells
- `Range`
  - add a `Range::range` fn which creates a sub-range from this range (clones or empty).

Closes #141

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "calamine"
-version = "0.15.2"
+version = "0.15.3"
 authors = ["Johann Tuffe <tafia973@gmail.com>"]
 repository = "https://github.com/tafia/calamine"
 documentation = "https://docs.rs/calamine"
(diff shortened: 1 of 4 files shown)


tafia:
@BillBarnhill perhaps you want to give it a look.

I have added a `Range::range` fn to query subranges. This is I think the most performant way to do it (except creating a new `DataTypeRef` like enum of course).

Let me know!

PR #148, opened 2019-04-08
improve deserializer

### LINKED ISSUE #115: Deserialize cells with numbers as strings
I am writing some software that deserializes a excel file as a `Vec<HashMap<String, Value>>` where Value is my own internal value type. I do this using the `RangeDeserializer`, with the headers setting set to `true`. However I am running into some problems when my header contains numbers, is deserialized as such, and I get an error about expecting string and getting a number. I guess a simple fix would to use my internal value type which can be deserialized as both a floating point and a string and then manually converting it to a string, however if the header is parsed as a boolean I would have problem there as well. I don't know if this is considered a bug or not, but is there any workaround other than using `Range::rows()` and manually implementing the deserializing.

tafia:
Close #115 

@birktj, @TimoFreiberg let me know if you can try it!

Without response, I'll merge it next week.

Thanks and sorry for the late fix!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,9 @@ failure = "0.1.3"
 failure_derive = "0.1.3"
 log = "0.4.6"
 serde = "1.0.80"
-quick-xml = "0.13.1"
+quick-xml = "0.14"
 zip = { version = "0.5.0", default-features = false, features = ["deflate"] }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


TimoFreiberg:
I'll try it this evening, thanks!

tafia:
I didn't see your edit great!

tafia:
published v0.15.4

Issue #153, opened 2019-09-09
Time is recognized as Float.


HiraokaTakuya:
[time.xlsx](https://github.com/tafia/calamine/files/3588635/time.xlsx)

time.xlsx is composed of one cell "00:15:00".
This cell recognized as Float(0.0104166666666667).
I expect as String("00:15:00").
Is this the intended behavior?
The test code is below.
```
#[test]
fn time() {
    setup();

    let path = format!("{}/tests/time.xlsx", env!("CARGO_MANIFEST_DIR"));
    let mut excel: Xlsx<_> = open_workbook(&path).unwrap();

    let range = excel.worksheet_range("Sheet1").unwrap().unwrap();
    range_eq!(range, [[String("00:15:00".to_string())]]);
}
```

tafia:
Yes, internal representations of date and times are float in excel.
Unfortunately calamine doesn't yest recognize that the number is actually a time and give a corresponding interpreted result.

HiraokaTakuya:
Oh, internal representations of times are float!
Thank you very much. I understood this behavior.
Do you have any plans to deal with cell formats in the future?

tafia:
I'd love to do it but I don't have lot of time to do it properly.
I'd accept any PR though 😄

HiraokaTakuya:
If I can improve it, I'll send PR.

tafia:
Just fyi, I've added a simple `dates` feature to try converting a `DataType` into a `NaiveDate(Time)`. There is no automatic detection as it is more complicated but hopefully it'll be convenient already.

HiraokaTakuya:
Thank you very much! This is very helpful. And I'm sorry I couldn't contribute to it.

Issue #116, opened 2018-02-18
Extracting dates from Excel


kper:
In my excel sheet, I've got a column with dates. Unfortunately, `calamine::DateType` has no variant for dates and when I handle a date as a string I get for `29.08.2012` -> `41181`. 

It would be nice to have built-in date conversion

kper:
Notes:

> Excel stores dates as sequential serial numbers so that they can be used in calculations. By default, January 1, 1900 is serial number 1, and January 1, 2008 is serial number 39448 because it is 39,447 days after January 1, 1900.

Source: https://support.office.com/en-us/article/datevalue-function-df8b07d4-7761-4a93-bc33-b7471bbff252

sts10:
I'm just learning Rust, but I got a work-around that might be a starting point? 

Basically I get an `i64` from any cells where I know ahead of time there will be a date (in other words, I haven't solved the detection logic mentioned above). These `i64`s are, as @kper's comment clued me to, the number of days since 1900. I then wrote the function below, using the [chrono crate](https://github.com/chronotope/chrono)'s NaiveDate type, to get a workable Date type. 

```Rust
extern crate chrono;
use chrono::NaiveDate; 
// ...
fn from_days_since_1900(days_since_1900: i64) -> NaiveDate {
    let d1900 = NaiveDate::from_ymd(1900, 1, 1);
    d1900 + Duration::days(days_since_1900 - 2)
}
```

I _believe_ that quirky `- 2` at the end is necessary for accuracy, perhaps to compensate for the Excel bug mentioned above? Anyway, hope this is helpful for other people who landed here looking to better handle dates, and/or an eventual PR!

tafia:
Thanks @sts10.

> I believe that quirky - 2 at the end is necessary for accuracy, perhaps to compensate for the Excel bug mentioned above?

On js-xlsx side, they seem to have a [different](https://github.com/SheetJS/js-xlsx/blob/3438923e5138f10de0aa70b35a8f56eedcfc320d/bits/20_jsutils.js#L34-L45) [logic](https://github.com/SheetJS/js-xlsx/blob/3438923e5138f10de0aa70b35a8f56eedcfc320d/bits/20_jsutils.js#L34-L45)

I don't have a lot of time to dig much deeper. I'd be happy to add a util function anyway so people can manually use it now, and hopefully it'll get converted automatically in the future.

sts10:
[...]

Issue #95, opened 2017-07-25
Expose typed data for dates and times


jrmuizel:
Currently they're just strings it would be nice if we could add support for exposing the type of these cells.

tafia:
Yes you're right. I have to dig into the documentation to parse this properly but I do not have much time at the moment, perhaps you'd like to give it a try?

yivo:
@tafia I am getting float value when trying to parse date-types cell.

I have included the example xlsx: 
[book.xlsx](https://github.com/tafia/calamine/files/1362871/book.xlsx)

dkkoval:
@tafia this functions from [xlsx-js](https://github.com/SheetJS/js-xlsx) might be useful:

* [`parse_date_code`](https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L204)
* [`write_date`](https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L290)
* [`fmt_is_date`](https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L654)
* [`read_date`](https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L1345)
* [`parseDate`](https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L1795)

Maybe just implement that in rust?

tafia:
Thanks!
I'm a little busy at the moment, I'll try to make it as fast as possible. Feel free to make a PR!

tafia:
> I can give it a try if you don't mind 😜

I am starting to look at it but this is far from trivial. I really don't mind if you look at it 😄 .

Instead of having a `calamine::DateTime` struct, I'd prefer, if possible having some existing item:
- either `std::time::Instant` 
- or `chrono::DateTime` (not sure if it is overkill or not)?

What do you think?

dkkoval:
I'd prefer unix timestamp (`i64`), cause it can be used in any time/date lib (e.g. [chrono](https://docs.rs/chrono/0.4.0/chrono/offset/trait.TimeZone.html#method.timestamp)).

tafia:
Looks good to me too!

nimtiazm:
Hi,
Has anyone worked on this ticket? If not completely, can you share what you've figured out so far? I mean what exactly needs to be done to be able to parse date/time columns?

tafia:
I had a look and having a perfect solution is not simple. Basically we need to parse the text format and decide upon it.

[...]

PR #156, opened 2019-10-11
Add `dates` feature

tafia:
This PR adds new fn to convert from `DataType` into `NaiveDate(Time)`s.
As it requires importing the whole chrono crate, this is behind a `dates` feature.


tafia:
related #153 #116 #95

PR #167, opened 2020-03-29
Add deserialize_with example to README

jamesmcm:
Added a simple example of using deserialize_with.


tafia:
Awesome!
Thanks for the PR and sorry for the delay!

PR #171, opened 2020-05-19
typo in fn description

bonusrk:
fn height has `Get column width` description now, so it goes to docs.rs docs also.


bonusrk:
I didn't find any contributions guide, so i am sorry if such simple prs are not needed.

tafia:
Thanks!

PR #172, opened 2020-06-05
Skip phonetic run

HiraokaTakuya:
calamine parse phonetic text just like ordinary text, but phonetic text is skipped typically.
See https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.phoneticproperties?view=openxml-2.8.1
I checked other library's behaviors, pandas (Python) and excelize (Go).
These libraries skip phonetic text.
Sample codes are here.
```python
#!/usr/bin/env python

import pandas as pd

input_book = pd.ExcelFile('rph.xlsx')
input_sheet_df = input_book.parse('Sheet1', header=None, index_col=None)

print(input_sheet_df[0][0]) # '課きく　毛こ'
```
```go
package main

import (
	"fmt"
	"github.com/360EntSecGroup-Skylar/excelize"
)

func main() {
	f, err := excelize.OpenFile("rph.xlsx")
	if err != nil {
		return
	}
	cell, err := f.GetCellValue("Sheet1", "A1")
	if err != nil {
		return
	}
	fmt.Println(cell) // "課きく　毛こ"
}
[rph.xlsx](https://github.com/tafia/calamine/files/4733297/rph.xlsx)
```


tafia:
Thanks!

PR #177, opened 2020-08-26
Fix XLS float parsing error

### LINKED ISSUE #176: Calamine incorrectly parses float as int in xls file
Hi,

When trying to decode the xls file embedded in [test.zip](https://github.com/tafia/calamine/files/5124336/test.zip) I get the result:
```
worksheet = Range {
    start: (
        0,
        0,
    ),
    end: (
        2,
        0,
    ),
    inner: [
        Int(
            0,
        ),
        Float(
            0.30000000000000004,
        ),
        Int(
            0,
        ),
    ],
}
```
However the file contains three cells with the values `0.5`, `0.3` and `0.7`. So apparently there is something funny going on with the cell parsing code (or am I misunderstanding something)?

The code I used to read the file:
```rust
let workbook = calamine::open_workbook_auto(&path)?;

let sheet_name = workbook.sheet_names().get(0).unwrap().to_owned();
let worksheet = workbook.worksheet_range(&sheet_name).unwrap();

dbg!(&worksheet);
```

birktj:
From testing it seems that numbers like `0.5` are sometimes encoded as ints with the `d100` flag set in `.xls` files. This patch converts these numbers to floats, as far as I understand this should be more correct? This fixes #176

### DIFF
--- a/src/xls.rs
+++ b/src/xls.rs
@@ -410,7 +410,12 @@ fn rk_num(rk: &[u8]) -> DataType {
     v[0] &= 0xFC;
     if is_int {
         let v = (read_slice::<i32>(&v[4..]) >> 2) as i64;
-        DataType::Int(if d100 { v / 100 } else { v })
+        if d100 && v % 100 != 0 {
+            DataType::Float(v as f64 / 100.0)
+        }
+        else {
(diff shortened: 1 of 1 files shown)


tafia:
Thanks!

PR #180, opened 2020-09-11
Correct MBSC to MBCS in vba.rs

Nopey:
Microsoft's multibyte character set was previously misspelt.
Here's an article on their website about it:
https://docs.microsoft.com/en-us/cpp/text/unicode-and-mbcs


tafia:
Thanks

PR #181, opened 2020-09-14
Update path formatting to Rust 2018

jamesmcm:
See https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html and https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html

The changes are pretty small overall, mostly just removing the `extern crate` declarations and adding a few anonymous lifetime markers on structs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,19 +58,9 @@
 //! ```
 #![deny(missing_docs)]
 
-extern crate byteorder;
-extern crate codepage;
-extern crate encoding_rs;
-extern crate quick_xml;
-#[macro_use]
-extern crate serde;
-extern crate zip;
-
-#[macro_use]
-extern crate log;
-
 #[macro_use]
 mod utils;
+
 mod auto;
 mod cfb;
 mod datatype;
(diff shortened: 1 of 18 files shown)


src/lib.rs
```diff
@@ -194,12 +184,7 @@ pub trait Reader: Sized {
     /// Get the nth worksheet. Shortcut for getting the nth
     /// sheet_name, then the corresponding worksheet.
     fn worksheet_range_at(&mut self, n: usize) -> Option<Result<Range<DataType>, Self::Error>> {
-        let name = if let Some(name) = self.sheet_names().get(n) {
-            name
-        } else {
-            return None;
-        }
-        .to_string();
+        let name = self.sheet_names().get(n)?.to_string();
```
tafia, line 187:
```suggestion
        let name = self.sheet_names().get(n)?.as_deref();
```

tafia (changes requested):
Thanks!
I've just made a simple suggestion to avoid allocating a new `String`

src/lib.rs
jamesmcm: Thanks, note this is nightly-only though: https://doc.rust-lang.org/std/result/enum.Result.html#method.as_deref

But so is the use of Bencher at the moment.

jamesmcm:
> I've just made a simple suggestion to avoid allocating a new `String`

Thanks, but I'm not sure this is possible since the method it is passed to uses &mut self, and the reference itself (name) implies an immutable reference over self.

I think we'd either have to move this inside the method that is taking &mut self (i.e. worksheet_range), or make that not take a mutable reference to self.

tafia:
Oh right!
Sorry for that. Could you revert the change?

jamesmcm:
Done 🙂

tafia:
Thanks!

Issue #182, opened 2020-09-17
[BUG] XLSB: Reader ignores formulas


kkirsanov:
Using example code https://github.com/tafia/calamine/blob/master/examples/excel_to_csv.rs on .xlsb with this content
`1|2|=sum(A1:B1)` produces csv `1;2`

 And when used on the XLSX  with the same content: `1;2;3`

tafia:
@kkirsanov can we consider this closed?

PR #183, opened 2020-09-18
Add the ability to read fomula values from XLSB

kkirsanov:
Resolving https://github.com/tafia/calamine/issues/182 with some help from https://github.com/egorvavilov


tafia:
Awesome!

PR #185, opened 2020-09-24
Add support for integral date types

dodomorandi:
There are [three](https://github.com/tafia/calamine/issues/153) [open](https://github.com/tafia/calamine/issues/116) [issues](https://github.com/tafia/calamine/issues/95) that are related to dates and maybe some of them can be closed. The discussion in #116 highlighted the case that an integral value could be used, but maybe something slipped through time ☺️ 

I would really appreciate a version bump with this feature, but unfortunately I think that it should be considered a breaking change 😞. Even if unintended, before this commit an integral value cannot be parsed as date, now it is a valid case instead. What do you think?

### DIFF
--- a/src/datatype.rs
+++ b/src/datatype.rs
@@ -115,13 +115,21 @@ impl DataType {
     /// Try converting data type into a datetime
     #[cfg(feature = "dates")]
     pub fn as_datetime(&self) -> Option<chrono::NaiveDateTime> {
-        self.get_float().and_then(|f| {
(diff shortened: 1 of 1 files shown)


tafia:
Thanks!

I don't think this particular change is a breaking change, I see it more as a bug fix (people want a date else they wouldn't call the function and before it was not working).

tafia:
bumped v0.16.2

PR #189, opened 2020-10-05
Ensure doctest functions actually run

robyoung:
It looks like doctest will only run a `main` function in an example. As a result all the `example` functions in the examples are not actually running.


tafia:
Thanks!

PR #197, opened 2020-12-18
Brute force file detection if extension is not known

### LINKED ISSUE #196: Cannot open files with no extension
Hello, running open_workbook_auto("filename_without_extension") will return the error:
Cannot open input file: Msg("Unknown extension")

Is there any way the library could automatically guess the file format?

gufoe:
Fixes #196


tafia:
Thanks

PR #198, opened 2021-01-06
Detect date/time formatted cells in XLSX

evilpie:
A lot of this logic is inspired by https://pypi.org/project/openpyxl/, which is also MIT licensed.
This handles built-in date formats and a some custom date formats that only use some common characters like HMS or :./.

This works with https://www.rki.de/DE/Content/InfAZ/N/Neuartiges_Coronavirus/Daten/Impfquotenmonitoring.xlsx?__blob=publicationFile, which uses a builtin date format. Libre Office always seems to produce a custom format so the testcase only uses that.


evilpie:
@tafia Hey! I am just wondering if you are still interested in this?

tafia:
Sorry for the late answer. Thanks a lot!

PR #201, opened 2021-02-01
Remove unsafe

### LINKED ISSUE #199: `Sectors::get` accesses unclaimed/uninitialized memory
Hello :crab: ,
we (Rust group @sslab-gatech) found a memory-safety/soundness issue in this crate while scanning Rust code on crates.io for potential vulnerabilities.

https://github.com/tafia/calamine/blob/2391d0c538734ee8c12e2b7610cc264999c04e78/src/cfb.rs#L243-L261

## Issue#1: Write to unclaimed memory.
`self.data.set_len(end)` is done without reserving extra memory for `self.data`.
After that, new data is written to the unclaimed memory via `r.read(&mut self.data[len..end])`.
**This seems to be a critical security error, as `r.read(&mut self.data[len..end])` can potentially write to memory that is actively occupied by another entity.**

## Issue#2: `read()` on uninitialized memory can cause undefined behavior

`r.read(&mut self.data[len..end])` can potentially pass uninitialized memory to user-provided `Read` implementation. This is unsound, because it allows safe Rust code to exhibit an undefined behavior (read from uninitialized memory).

[This part](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) from the `Read` trait documentation explains the issue:

> It is your responsibility to make sure that `buf` is initialized before calling `read`. Calling read with an uninitialized `buf` (of the kind one obtains via `MaybeUninit<T>`) is not safe, and can lead to undefined behavior.

## Suggested Fix
It would be safe to zero-initialize the newly allocated `u8` buffer before `read()`
(probably via `self.data.resize(end - len, 0)`),
in order to prevent user-provided `Read` from accessing old contents of the newly allocated heap memory.

Thank you for checking out this issue :+1:

dodomorandi:
This PR fixes the following issues:
* Issue #199
* Unsound implementation of `to_u32`
* Unsound implementation of `read_slice`

The unsoundness of the last two issues is reproducible after commit https://github.com/tafia/calamine/commit/d06428a7a4a34ffe6447b0718f6fe27b5da2f595 using miri.
`to_u32` has been modified in order to return an iterator of `u32`s.
Similarly, `read_slice` has been removed and an implementation has been written for each type. Unfortunately, with the current limitations of the trait system, it is not possible to create a generalized version because of the lifetime of the output iterator. I noticed that most of the time only the first value from the iterator is really needed. Moreover, to maintain the current behavior, it is assumed that the value is always obtained, therefore `unwrap` has been used, but a proper error handling is recommended.

I also recommend at least a patch version bump after this PR is merged.
Fixes #199 

*EDIT*: I forgot to say that I introduced usage of `TryFrom`, therefore I could have bumped MSRV to 1.34,

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -167,7 +167,7 @@ impl Header {
         f.read_exact(&mut buf).map_err(CfbError::Io)?;
 
         // check ole signature
-        if read_slice::<u64>(buf.as_ref()) != 0xE11A_B1A1_E011_CFD0 {
+        if read_slice_u64(buf.as_ref()).next() != Some(0xE11A_B1A1_E011_CFD0) {
             return Err(CfbError::Ole);
         }
 
@@ -291,15 +291,20 @@ struct Directory {
 
 impl Directory {
     fn from_slice(buf: &[u8], sector_size: usize) -> Directory {
+        use std::convert::TryFrom;
+
         let mut name = UTF_16LE.decode(&buf[..64]).0.into_owned();
         if let Some(l) = name.as_bytes().iter().position(|b| *b == 0) {
             name.truncate(l);
         }
         let start = read_u32(&buf[116..120]);
         let len = if sector_size == 512 {
-            read_slice::<u32>(&buf[120..124]) as usize
+            read_slice_u32(&buf[120..124]).next().unwrap() as usize
         } else {
-            read_slice::<u64>(&buf[120..128]) as usize
+            read_slice_u64(&buf[120..128])
+                .next()
+                .and_then(|x| usize::try_from(x).ok())
+                .unwrap()
         };
 
         Directory { start, len, name }
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -83,7 +83,7 @@ impl Cfb {
         debug!("load difat");
         let mut sector_id = h.difat_start;
         while sector_id < RESERVED_SECTORS {
-            difat.extend_from_slice(to_u32(sectors.get(sector_id, reader)?));
+            difat.extend(to_u32(sectors.get(sector_id, reader)?));
             sector_id = difat.pop().unwrap(); //TODO: check if in infinite loop
         }
 
@@ -91,7 +91,7 @@ impl Cfb {
         debug!("load fat");
         let mut fats = Vec::with_capacity(h.fat_len);
         for id in difat.into_iter().filter(|id| *id != FREESECT) {
-            fats.extend_from_slice(to_u32(sectors.get(id, reader)?));
+            fats.extend(to_u32(sectors.get(id, reader)?));
         }
 
         // get the list of directory sectors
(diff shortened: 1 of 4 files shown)


tafia:
Awesome!
Would you mind running the benches and see the impact?

dodomorandi:
Sure! These are comparative benches:

### Master
```
test bench_ods  ... bench:  15,636,817 ns/iter (+/- 1,539,111)
test bench_xls  ... bench:      42,027 ns/iter (+/- 899)
test bench_xlsb ... bench:     175,716 ns/iter (+/- 2,976)
test bench_xlsx ... bench:     219,132 ns/iter (+/- 3,343)
```

### e88c6ff22e0d1a7b0733d9b90fadc4722763d7d5
```
test bench_ods  ... bench:  15,225,132 ns/iter (+/- 331,808)
test bench_xls  ... bench:      41,958 ns/iter (+/- 3,530)
test bench_xlsb ... bench:     174,162 ns/iter (+/- 1,588)
test bench_xlsx ... bench:     217,257 ns/iter (+/- 1,482)
```

### 8ff72180167e85048cd4e15bf2301a2d9e7338cc
```
test bench_ods  ... bench:  15,254,287 ns/iter (+/- 525,769)
test bench_xls  ... bench:      43,444 ns/iter (+/- 983)
test bench_xlsb ... bench:     175,575 ns/iter (+/- 1,003)
test bench_xlsx ... bench:     217,667 ns/iter (+/- 11,103)
```

### a638e6f285eabcf3b3c8696707a89f86d16d947c
```
test bench_ods  ... bench:  15,404,189 ns/iter (+/- 145,735)
test bench_xls  ... bench:      43,676 ns/iter (+/- 893)
test bench_xlsb ... bench:     176,741 ns/iter (+/- 1,228)
test bench_xlsx ... bench:     217,512 ns/iter (+/- 971)
```
---

As you can see there is a small significative regression with the new implementation of `to_u32`. I _underestimated_ the cost of using `chunks` instead of `chunks_exact`. I just tested it, using `chunks_exact` almost recovers the regression. I am going to make a further commit with this change. :blush:

dodomorandi:
The new commit:

### b42ac8329a4ecf437f247189ae06235c62966f86
```
test bench_ods  ... bench:  15,402,172 ns/iter (+/- 389,565)
test bench_xls  ... bench:      42,437 ns/iter (+/- 1,341)
test bench_xlsb ... bench:     176,646 ns/iter (+/- 2,866)
test bench_xlsx ... bench:     220,921 ns/iter (+/- 1,574)
```

tafia:
Thanks, would you mind comparing with master on your system (else it isn't really that useful). Ideally with https://crates.io/crates/cargo-benchcmp.

dodomorandi:
This is with cargo-benchcmp:
```
 name        master.bench ns/iter  remove-unsafe.bench ns/iter  diff ns/iter  diff %  speedup
 bench_ods   15,678,339            15,555,000                       -123,339  -0.79%   x 1.01
 bench_xls   41,896                42,529                                633   1.51%   x 0.99
 bench_xlsb  174,118               177,445                             3,327   1.91%   x 0.98
 bench_xlsx  216,852               218,900                             2,048   0.94%   x 0.99
```

tafia:
Awesome, thanks!

PR #207, opened 2021-02-05
Add big endian test to travis

dodomorandi:
While I was making my last PR, I noticed that the data decoding was assuming a little endian machine.

I just added a big endian arch to travis in order to test both major endianness models. This obviously makes tests fail, but at least it is a good starting point to fix things.


tafia:
Thanks!

PR #208, opened 2021-02-06
Improve conversions from raw data to primitives

dodomorandi:
This is a follow-up of #201, in order to fix big endian issues (it can be tested with #207) and simplify code.

Changes:
* Use `from_le_bytes` instead of `from_ne_bytes`. The original PR intended to maintain the behavior of the original code, restricting its purpose to remove unsafe and UBs. This PR, instead, is meant to improve support to big endian architectures.
* Changed data-to-primitive abstractions in order to return a single value instead of an iterator. **In theory** previous code should be zero overhead (if the optimizer was able to throw away a bunch of stuff), now it is surely simpler for both coders and optimizer. I do not expect any significant performance difference, but simpler is obviously better in this case.
* Use [`impl<T, const N: usize> TryFrom<&[T]> for [T; N]]`](https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3C%26%27_%20%5BT%5D%3E) to simplify code. I incorrectly remembered this to be introduced in recent versions, but it's available since 1.34, so it is worth using it.

Performance difference from `master` is not significant:
```
 name        master ns/iter  improve-primitive-conversions ns/iter  diff ns/iter  diff %  speedup
 bench_ods   15,187,084      15,553,032                                  365,948   2.41%   x 0.98
 bench_xls   42,718          42,697                                          -21  -0.05%   x 1.00
 bench_xlsb  169,298         169,195                                        -103  -0.06%   x 1.00
 bench_xlsx  246,964         249,108                                       2,144   0.87%   x 0.99
```

Relates to #207

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -2,6 +2,7 @@
 
 use std::borrow::Cow;
 use std::cmp::min;
+use std::convert::TryInto;
 use std::io::Read;
 
 use log::debug;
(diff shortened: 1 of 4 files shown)


tafia:
Thanks

Issue #205, opened 2021-02-05
[FEATURE] iterate through worksheets?


Zerowalker:
Currently if you want to iterate through all worksheets then you have to do something like:

`   let mut excel: Xlsx<_> = open_workbook("excel/test.xlsx").unwrap();
    let names = excel.sheet_names().to_owned();
    for name in names.iter() {
        if let Some(Ok(r)) = excel.worksheet_range(name) {
         .....
         }`

I am wondering if it's add an iterator for this directly, perhaps even returning the name at the same time,
so something like.

`   let mut excel: Xlsx<_> = open_workbook("excel/test.xlsx").unwrap();
    for (name, sheet) in excel.worksheets.iter() {
     ......
    }`

Fuzen-py:
This seems like a common enough use case to warrant a shortcut method. I have opened a draft. Any contributions / feedback are welcome as I am new to this project.

Fuzen-py:
Looks like this was also request in #165.

PR #209, opened 2021-02-15
Feature: Iterate through all worksheets

Fuzen-py:
Allow users to quickly iterate through all worksheets addressing #205.


src/lib.rs
```diff
@@ -159,6 +159,19 @@ pub trait Reader: Sized {
     fn metadata(&self) -> &Metadata;
     /// Read worksheet data in corresponding worksheet path
     fn worksheet_range(&mut self, name: &str) -> Option<Result<Range<DataType>, Self::Error>>;
+
+    /// Fetch all worksheets
+    fn worksheets(&mut self) -> Vec<(String, Range<DataType>)> {
```
Fuzen-py, line 164: I have opted to use a `Vec` over a range, as there is no real reason to request a sheet by name, `worksheet_range` exists for this reason.

tafia:
LGTM

PR #211, opened 2021-02-20
Fix build failure on Debian GNU/Linux 10

striezel:
**Short summary:**
calamine 0.17.0 does not build with the default Rust compiler that comes with Debian GNU/Linux 10, but version 0.16.2 did build just fine on that system. This pull request makes some adjustments so that it builds again and Debian users can enjoy the newer version, too.

**Longer explanation:**
[Debian](https://www.debian.org/) is a Linux distribution, currently ranked 6th most-popular according to [DistroWatch](https://distrowatch.com/table.php?distribution=debian). Like other Linux distributions it comes with various prepackaged software, and among that software also is the Rust compiler. However, the current stable release of Debian comes with Rust 1.41.1, which is [almost a year old](https://blog.rust-lang.org/2020/02/27/Rust-1.41.1.html). Therefore, it does not support some of the newer Rust features from the current stable branch of Rust. The two main "problematic" newer features are:

* the [`matches!` macro](https://doc.rust-lang.org/std/macro.matches.html), introduced in Rust 1.42.0, and
* std::path::Path::strip_prefix() that uses a library feature that was still unstable when Rust 1.41.1 was released.

The pull request replaces both of them with compatible versions that also work with Rust 1.41. I tried to use the corresponding code snippets that were used in those places in calamine 0.16.2 as replacements, so that the functionality should not be impacted. (The tests still pass.)

Furthermore, I took the liberty to add a configuration file for Clippy, `.clippy.toml`, which sets the Minimum Supported Rust Version (MSRV) to 1.41.0. That way Clippy does not warn about using the `matches!` macro anymore. I guess that was the reason why someone replaced the match expressions with the macro in the first place.

_I am aware that these changes may be a bit intrusive,_ so if you decide not to merge these changes, I'm fine with that, too. Because what this pull request also does as a side effect is setting the MSRV of calamine to 1.41.0. I was unable to find an explicit statement what calamine's current MSRV is, so I am not sure whether this is OK or not.


tafia:
Thanks!

I can't guarantee that the MSRV will always be in line with what Debian is providing but for now, this particular macro doesn't bring anything but a shorter syntax.

striezel:
Thanks for merging, that certainly helps. :+1:

I would also appreciate it, if you could publish the current master as a new version to crates.io.  That way people pulling the latest version from crates.io get a version that also works with older Rust compilers which means that it can be used by a broader audience.

As for the MSRV guarantee: I am aware of that. The usual solution for this is that the maintainers of a Rust library commit to support a specific MSRV and only bump the MSRV in certain pre-defined intervals, e. g. once a year or after a certain number of Rust releases so that the last N releases of Rust are always supported. But that is a discussion for another issue / pull request.

tafia:
Doing it tonight (asia time)!

PR #213, opened 2021-03-10
Add xlsx table parsing

### LINKED ISSUE #163: Add read table support
Would it be possible to add Excel table support?  The ability to identify cell ranges that are part of tables.

I'm guessing it would similar to java/apache/poi
```
XSSFWorkbook wb = new XSSFWorkbook(inp);
for (int iSheet = 0; iSheet < wb.getNumberOfSheets(); iSheet++) {
    XSSFSheet sheet = wb.getSheetAt(iSheet);
    TreeMap<String, XSSFTable> orderTable = new TreeMap<>();
    for (XSSFTable xlTable : wb.getSheetAt(iSheet).getTables()) {
        CellReference endCell = xlTable.getEndCellReference();
        CellReference startCell = xlTable.getStartCellReference();

        for (iRow = startCell.getRow(); iRow <= endCell.getRow(); iRow++) {
            Row row = sheet.getRow(iRow);
            for (iCol = startCell.getCol(); iCol <= endCell.getCol(); iCol++) {
            	... do the things ...
	    }
    }
    ... do the things ...
}
```

jfcha:
This PR adds reading table support for the xlsx file format. This partial resolves #163 

Right now the design requires users to call load_tables so that there is no performance impact on people who don't want/need this functionality but it could easily be adjusted to load the tables when loading the initial file metadata. This functionality could also hide behind a feature flag. 

The API is pretty simple:

```rust
    let mut xls: Xlsx<_> = open_workbook(&path).unwrap();
    xls.load_tables().unwrap();
    let table_names = xls.table_names();
    assert_eq!(table_names[0], "Temperature");
    let table = xls
        .table_by_name("Temperature")
        .expect("Table should exists")
        .expect("Parsing table worksheet should not error");
    assert_eq!(table.name(), "Temperature");
    assert_eq!(table.columns()[0], "label");
    assert_eq!(table.columns()[1], "value");
    assert_eq!(
        table.data().get((0, 0)),
        Some(&String("celsius".to_owned()))
    );
```
but I'm happy to adjust the API as needed. I needed this functionality for a project so I thought I'd share it.


src/xlsx.rs
```diff
@@ -360,6 +362,224 @@ impl<RS: Read + Seek> Xlsx<RS> {
...
+                            }
+                        }
+                    }
+                    Ok(Event::End(ref e)) if e.local_name() == b"Relationships" => break,
+                    Ok(Event::Eof) => return Err(XlsxError::XmlEof("Relationships")),
+                    Err(e) => return Err(XlsxError::Xml(e)),
+                    _ => (),
+                }
+            }
+            drop(xml); // this drop shouldn't be necessary, but the borrow checker complains otherwise
```
tafia, line 425: I'd prefer putting the whole thing in a block `{...}` if you don't mind.

tafia (changes requested):
Awesome!

tafia (approved):

tafia:
Thanks!!

PR #216, opened 2021-04-06
Check For Existing Tables When Collecting

WCarrollSTO:
When loading tables, only the final sheet containing tables is returned.

EX: 
- Sheet 1 has a table
- Sheet 2 has two tables
- Sheet 3 has none.

Result:
- 2 tables are found (just Sheet 2)

This PR appends the tables if any have been found so far, otherwise it will still replace.

The test suite has been updated to check for this.


tafia:
Thanks!

PR #223, opened 2021-06-22
Add deserialize_enum implementation

markcatley:


tafia:
Thanks!

TeXitoi:
Hi! can we have a release? I'm particularly interested by this evolution. The git version works for me, but I'd prefer to use a published crate.

PR #232, opened 2021-09-08
include ":" when detecting datetimes

sezna:
A datetime format in an xlsx often includes a `:`, e.g. `h:m`. This change includes `:` in the list of possible formatter characters.


tafia:
Thanks!

PR #234, opened 2021-11-10
Enable .xlsx parsing for namespaced definedName values

dhbradshaw:
This continues using a pattern for parsing namespaced xlsx files introduced in pr
https://github.com/tafia/calamine/pull/73 and thereby allows parsing of namespaced .xlsx
files with definedName values.

`xml.read_text` requires the fully namespaced name.  Since that isn't
always equal to the local_name, the pattern introduced has been to
use `e.local_name` for making comparisons to the raw value of interest and then `e.name()`
instead of that raw value for xml.read_text .


tafia:
Thanks

Issue #219, opened 2021-04-18
Error on open_workbook for attached file


Edgeworth:
The following error occurs for the file I've attached, upon trying to open it using open_workbook:

"Continued record too short while reading extended string"

This happens on calamine 0.18.0 but not 0.17.0.

[murc_2021.xls](https://github.com/tafia/calamine/files/6330387/murc_2021.xls)

tafia:
Thanks for reporting it, I'll try to have a look but I don't have lot of time at the moment.

Edgeworth:
Hey, no worries. I'm just using 0.17.0. It's non-urgent for me at least.

Edgeworth:
I confirmed that this is no longer an issue, so closing this.

PR #236, opened 2021-11-22
fix `open_workbook()` for xls may panic or failed.

visig9:
Related issue: #219 

This bug happen because `xls::read_rich_extended_string()` not skip `ExtRst` data block correctly with the size indicated by `cbExtRst` field. So `open_workbook()` may return `Err(XlsError::ContinueRecordTooShort)` or even panic by index out of bound problem.

See also: <https://docs.microsoft.com/en-us/openspecs/office_file_formats/ms-xls/173d9f51-e5d3-43da-8de2-be7f22e119b9>

# Testcase

The new testcase not use original #219 attachment file, but create by myself. The steps as follows:

1. Open MS Excel (Mac Version, ver: "16.55 (21111400)") 
2. File -> New
3. Write "a" in first field.
4. Use "save as" function to save as xls format.


tafia:
Awesome thanks!!

PR #254, opened 2022-03-31
Use iterator whenever possible instead of manual iteration

hbina:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,6 +205,9 @@ where
 
 /// A trait to constrain cells
 pub trait CellType: Default + Clone + PartialEq {}
+
+// Blanket implementation of `CellType` for any `T` that satisfy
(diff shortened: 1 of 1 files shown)


JustForFun88:
> Signed-off-by: Hanif Ariffin [hanif.ariffin.4326@gmail.com](mailto:hanif.ariffin.4326@gmail.com)

Xm... previous implementation use only 3 checks per iteration:
1. That we inside cells range;
2. Checking col_start;
3. Checking col_end.

The new one will be use 4 checks, because now you check col_start and col_end separately, and that means that checking cells range will also be done twice.
This PR slow down the performance and compilation time.

tafia:
I agree with @JustForFun88. I don't really see that PR as necessary.
Closing it for now, @hbina feel free to reopen it if you think we're wrong.

PR #256, opened 2022-03-31
Implement API to open workbook from bytes

hbina:


upachler:
I'm also interested in this feature, for similar reasons (I want to write an asset loader that can read Excel sheets for bevy, a game engine written in Rust also supporting web assembly). 
In general I think it is good practice to have loader libraries not being dependent on the file system - because in the end, loaders don't consume files, they consume bytes..

tafia:
Thanks a lot and sorry for the time it took to merge.

PR #257, opened 2022-03-31
Applied some suggestions by Intellij

hbina:
1. Fix some small spelling error
2. Reorder the trait functions
3. Use usize::MAX instead of std::usize::MAX which is due for deprecation


tafia:
Thanks!

PR #267, opened 2022-08-01
Do not let chrono pull in time 0.1.44

dusdanig:
This solves RUSTSEC-2020-0071 when dates feature is enabled. Calamine
does not use the current time, so it does not need the dependency on the
time crate (and it's security concern).

Bumped the version of chrono to 0.4.19.

This solves: https://github.com/rustsec/advisory-db/blob/main/crates/time/RUSTSEC-2020-0071.md


tafia:
Thanks!

PR #269, opened 2022-09-20
Add DateFormat to DataType::Datetime

econaxis:
For my use case, I need to recover the original string as Excel displays it. To do this, I need to know the date format (e.g. "Jan 12 2003" vs "01-12-03"). 

This PR adds a format value to DataType::Datetime and functions to display the date given a date float and a format specifier.

I will add tests for the new stuff after you give the go-ahead to add this feature to the crate.

### DIFF
--- a/src/xlsx.rs
+++ b/src/xlsx.rs
@@ -13,6 +13,7 @@ use zip::result::ZipError;
 
 use crate::vba::VbaProject;
 use crate::{Cell, CellErrorType, DataType, Metadata, Range, Reader, Table};
+use crate::datatype::DateFormat;
 
 type XlsReader<'a> = XmlReader<BufReader<ZipFile<'a>>>;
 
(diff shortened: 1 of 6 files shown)


src/xlsx.rs
```diff
@@ -234,10 +235,11 @@ impl<RS: Read + Seek> Xlsx<RS> {
                                     .map_or(CellFormat::Other, |a| {
                                         match number_formats.get(&*a.value) {
                                             Some(fmt) if is_custom_date_format(fmt) => {
-                                                CellFormat::Date
+                                                CellFormat::Date(DateFormat::Unknown)
                                             }
-                                            None if is_builtin_date_format_id(&a.value) => {
-                                                CellFormat::Date
+                                            None if is_builtin_date_format_id(&a.value).is_some() => {
+                                                let date_format = is_builtin_date_format_id(&a.value).unwrap();
```
KillTheMule, line 241: This should return an error instead of panicking.

tafia:
There are conflicts in this PR and date format detection has been implemented now.
There is no actual formatting though, feel free to open another PR with the current master.
Thanks

PR #272, opened 2022-10-03
crate: add keywords for excel extensions

kov:
I almost missed this great crate because I search for crates that had the xlsx keyword =)


tafia (approved):

tafia:
Thanks

Issue #262, opened 2022-06-09
Worksheets not sorted when parsing an xls file


danipardo:
If I open an xls format file, like in

```
    let workbook: Xls<_> = open_workbook("sample.xls").unwrap();
```
The vec returned by workbook.worksheets() seems to be in random order. I would expect them to be in order, otherwise we will have to rely on worksheet name to find the apropiate one. In my case, the user uploads an excel file, and I know the data is in the first worksheet, whatever name it has.

cnpryer:
I'm interested in preserving order here too, but it looks like it's stored as a hashmap internally. So this is probably where order isn't preserved.

On a side note, I'm curious if this project is still maintained. 🤔

Fuzen-py:
There has not been any updates from Tafia for quite some time. I sent the author an email. In the meantime, I will look into a solution for you, it may be adding a separate method ordered worksheets.

cnpryer:
I'm open to contributing to someone's fork if there's enough support. Otherwise I'll probably just queue up PRs here and point at my own.

Fuzen-py:
It looks like the worksheet's metadata is sorted, not really a great solution but may work for you?

```rust 
let workbook: Xls<_> = open_workbook("sample.xls").unwrap();
let mut sheets = workbook.worksheets();
let metadata = workbook.metadata();
let sheet_names = &metadata.sheets;

sheets.sort_by(|left, right| {
    let left_pos = sheet_names.iter().position(|v| left.0.eq(v));
    let right_pos = sheet_names.iter().position(|v| right.0.eq(v));

     left_pos.cmp(&right_pos)
});
```

cnpryer:
I'll check it out. Thanks!

Fuzen-py:
For @danipardo case, if you know the position `.sheet_names().get(2)` should give you the results you want as well if you fetch the range with `worksheet_range`. Again, this is under the assumption the metadata is in order. If not, we should probably store positioning there.

danipardo:
Yes, workbook.sheet_names() seems to be sorted, so I guess we can work around with that. Thanks!

danipardo:
Hi again, @Fuzen-py . Did you finally get a reply from @tafia about calamine project status? It's a pity the project is stuck, given how often people need to parse excel files.

Fuzen-py:
@danipardo I have not heard back from the maintainer. I'm open to either forking this project or contributing to another one.

danipardo:
[...]

PR #275, opened 2022-10-19
use BTreeMap instead of HashMaps

### LINKED ISSUE #271: cannot read value some time at the same cell
```rust
pub fn test_xls() -> Result<(), Error> {
      use std::fmt::Write;
      let mut count = 0;
      let mut buffer = String::new();
      loop {
          let path = format!("{}/testfile.xls", env!("CARGO_MANIFEST_DIR"));
          let mut workbook = open_workbook_auto(path)?;
          let v = workbook.worksheets();
          let (_sheetname, range) = v.first().ok_or(Error::Msg("bad format"))?;
          let value = range.get( (0, 1) )
          .map(|s| s.to_string());
          write!(&mut buffer, "{}: {:?}\r\n", count, value).unwrap();
          count += 1;
          if count > 20 {
              break;
          }
      }
      std::fs::write("output.txt", buffer.as_bytes()).unwrap();
      Ok(())
}
```
output.txt:

0: Some("yyy_name")
1: Some("yyy_name")
2: Some("yyy_name")
3: None
4: None
5: None
6: Some("yyy_name")
7: None
8: Some("yyy_name")
9: None
10: Some("yyy_name")
11: Some("yyy_name")
12: Some("yyy_name")
13: Some("yyy_name")
14: Some("yyy_name")
15: None
16: Some("yyy_name")
17: None
18: Some("yyy_name")
19: None
20: None

[dependencies]
calamine = {git = "https://github.com/tafia/calamine.git"}

[testfile.xls](https://github.com/tafia/calamine/files/9630527/testfile.xls)

tafia:
Close #271 

HashMap almost guarantee that the items order will change upon successive runs. 
Use BTreeMap instead to have reproducible runs. Given the number of expected items and the simplicity of the keys (Strings mostly) I don't expect any particular bad impact.


tafia:
Should also close #262

PR #278, opened 2022-10-28
Add a check that the number format returned isn't an empty string

### LINKED ISSUE #215: Date detection seems to be too agressive
**Expected Behavior**
Integer values should still be handled as integers

**Actual Behavior**
Cells with integer values in the [attached](https://github.com/tafia/calamine/files/6186287/Profile.xlsx) XLSX file get treated as date times.

**Effected Versions**
0.17 and 0.18

**Details**
I recently updated from 0.16 to 0.18, although 0.17 is affected as well after testing. I process a XLSX document to perform some automatic code generation for my [fitparser](https://github.com/stadelmanma/fitparse-rs) crate (I've attached the file: [Profile.xlsx](https://github.com/tafia/calamine/files/6186287/Profile.xlsx)).  After merging in feature #198 it seems to be treating the cells with an integer value in Sheet "Types", Column D as date times. This also effects Column B in the "Messages" sheet.

Initially I tried simply coercing the values to an i64 which works _most_ of the time but seems to fail on larger values, which I assume must not be valid dates.

Rust Versions:
rustc 1.49.0 (e1884a8e3 2020-12-29)
cargo 1.49.0 (d00d64df9 2020-12-05)

stadelmanma:
Fixes #215 

In the custom date detection code `format.bytes().all(|c| b"mdyMDYhsHS-/.: \\".contains(&c))` matches the empty string. The excel files I'm processing might have issues to have an empty number format in them to begin with but I don't think we want those treated as date strings either way.

### DIFF
--- a/src/xlsx.rs
+++ b/src/xlsx.rs
@@ -221,7 +221,9 @@ impl<RS: Read + Seek> Xlsx<RS> {
                                     _ => (),
                                 }
                             }
-                            number_formats.insert(id, format);
+                            if !format.is_empty() {
+                                number_formats.insert(id, format);
+                            }
                         }
                         Ok(Event::End(ref e)) if e.local_name().as_ref() == b"numFmts" => break,
                         Ok(Event::Eof) => return Err(XlsxError::XmlEof("numFmts")),


tafia:
Thanks!

PR #282, opened 2022-11-07
Add cargo-fuzz fuzzing harness

MJVL:
Added a fuzzing harness such that [cargo fuzz](https://github.com/rust-fuzz/cargo-fuzz) can be used to help identify bugs or security issues automatically.

Fuzzing can be ran with `cargo fuzz run fuzz_all`

While not inclusive of all behavior, the current harness tests:
* Opening a workbook
* Opening worksheets and inspecting ranges
* Looking for vba projects, modules, and referernces
* Checking worksheet formulas

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target
 Cargo.lock
 *.bk
 .vim
+fuzz.xlsx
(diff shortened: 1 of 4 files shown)


tafia:
Thanks!

PR #285, opened 2023-01-06
Fix having row 0 in dimensions

roblabla:
I have an XLSX document that was automatically generated by some tool that calamine chokes on. A quick investigation showed that its sheet had the dimension defined as `<dimension ref="A0:AZ48943" />`. This would cause either a panic when calamine was compiled in debug mode, or very odd behavior in calamine in release mode (usually ending with panics).

This is obviously broken, but both Excel and LibreOffice seem to handle it just fine. So I did a sort of quick-and-dirty fix so the XLSX could be parsed by calamine.


tafia:
Thanks!

PR #289, opened 2023-02-03
ODS: handle text:s tags

### LINKED ISSUE #225: [BUG] ODS: Reader doesn't handle text:s tags
When parsing ODS files that have multiple spaces within a cell, duplicate spaces are removed because they are stored as `<text:s />` (which may have a count parameter)

The behaviour of stripping extra spaces is not observed when reading from Xls or Xlsx files.

Would expect consistent behaviour in not stripping extra spaces in cell content.

The test suite for "special_cells.ods" seems to explicitly test that the extra spaces are skipped.

I'm developing a program that matches certain cells by an exact string match and supports Ods, Xls and Xlsx files, only Ods is causing mismatches due to extra spaces being removed.

[test.ods](https://github.com/tafia/calamine/files/6706950/test.ods)
[test.xls](https://github.com/tafia/calamine/files/6706975/test.xls)
[test.xlsx](https://github.com/tafia/calamine/files/6706976/test.xlsx)

dimastbk:
This PR should close #225.


tafia:
Thanks!

PR #293, opened 2023-02-14
Dependency: update quick-xml to 0.27

WindSoilder:


WindSoilder:
Hi, I found out that the dependency `quick-xml` is not up to date, which may cause some applications (like nushell) to compile it multiple times. It would be nice if quick-xml was updated and released without breaking semver compatibility in version numbers.  Then our echo system will benefit from it too :) What do you think?

tafia:
Thanks!

Issue #292, opened 2023-02-13
No DateTime support in XLSB files


ketbra:
I'm getting different behavior when reading XLSB files as opposed to XLSX files.  The cells always come out as DataType(f64) instead of DateTime(f64).  I'm guessing this is just not implemented?  Looking in xlsb.rs I don't see any styles read in the way they are for XLSX files from xl/styles.xml in xlsx.rs.  I'm guessing date support requires reading xl/styles.bin from inside of XLSB files.  Am I correct that DateTime support is just not implemented for XLSB files?  I'm happy to take a look, but I wanted to see if you agree that it's just an issue of reading styles form xl/styles.bin and then applying similar logic for determining datetime styles as exists in xlsx.rs?

tafia:
As explained in other issues, unfortunately this is how dates are encoded.
You can manually convert them using the date features and `DataType::as_date` function.

https://github.com/tafia/calamine/blob/75de8f2bda047df89af7360a651dbcb10480d426/tests/test.rs#L750

ketbra:
> As explained in other issues, unfortunately this is how dates are encoded. You can manually convert them using the date features and `DataType::as_date` function.
> 
> https://github.com/tafia/calamine/blob/75de8f2bda047df89af7360a651dbcb10480d426/tests/test.rs#L750

Thanks tafia.  I think I did a poor job explaining the issue.  Code is clearer than words, so I implemented my proposed fix, a test demonstrating the desired behavior, and created a pull request.  I tried to keep the style of the existing code, but please let me know if you want anything changed!

PR #296, opened 2023-03-10
Cells are now returned as DataType::DateTime enums where appropriate …

ketbra:
#292 I implemented the approach I described, and now XLSB returns DataType::DateTime enums in a manner consistent with the XLSX parsing.  I added a new test, date_xlsb, which now passes but would not have before.  This test is a duplicate of date_xlsx except that I opened the date.xlsx test file and resaved it as date.xlsb.


tafia (approved):
Awesome, Thanks!

PR #299, opened 2023-03-29
XLS: support dates

dimastbk:
Hi! This PR adds date support to XLS. Yes, copying `is_custom_date_format`/`is_builtin_date_format_id` is not good, but, I think, it can be refactored in future PRs, after merge this and #296 (see also #303).

I also added support for 1904 date, but that is just hack now. Maybe add a flag/enum in DataType::DateTime()? This can also be used in xlsx/xlsb.


tafia:
Awesome

PR #303, opened 2023-04-30
Improve date format parsing

dimastbk:
Improve date format parsing, added in #198. This logic ported from openpyxl, which is under MIT License.
It also passes almost all tests from SheetJS, [except one](https://github.com/SheetJS/sheetjs/blob/5b4806bff01a9f4289b4c04e8a60a6c4c52fe93a/packages/ssf/test/is_date.json#L223).


tafia (changes requested):
Importing the whole regex for this use case seem overkill!

At the end of the day, if I understand the code properly, the only thing that matters to be considered a "datetime" pattern is :
- clean up all "special" parts (locale code, quoted text etc ...)
- check if there is one of "dmhysDMHYS" character

So, here is a new proposal which I believe works the same (passes your test at least) but is 
- simpler to understand (and tweak if necessary)
- properly commented
- zero copy
- probably faster

```rs
fn is_custom_date_format(format: &str) -> bool {
    format
        // take the first format only
        .split(';')
        .take(1)
        // remove anything in quotes
        .flat_map(|f| f.split('"').step_by(2))
        // remove anything in bracket (except datetime chars)
        .flat_map(|f| f.split('['))
        .map(|f| {
            if let Some((inside, rest)) = f.split_once(']') {
                if !matches!(
                    inside.as_bytes(),
                    &[b'h' | b'm' | b's' | b'H' | b'M' | b'S']
                        | &[b'h' | b'H', b'h' | b'H']
                        | &[b'm' | b'M', b'm' | b'M']
                        | &[b's' | b'S', b's' | b'S']
                ) {
                    return rest;
                }
            }
            f
        })
        // remove all escaped characters
        .flat_map(|f| {
            f.split(|c| c == '\\' || c == '_')
                .enumerate()
                .map(|(i, f)| if i == 0 { f } else { &f[1..] })
        })
        .flat_map(|f| f.chars())
        .any(|c| matches!(c, 'd' | 'm' | 'h' | 'y' | 's' | 'D' | 'M' | 'H' | 'Y' | 'S'))
}
```

tafia:
A variant more imperative but probably faster (read the characters only once)

```rs
fn is_datetime(s: &str) -> bool {
    let mut escaped = false;
    let mut is_quote = false;
    let mut brackets = 0u8;
    let mut prev = ' ';
    let mut hms = false;
    for s in s.chars() {
        match (s, escaped, is_quote, brackets) {
            (_, true, ..) => escaped = false, // if escaped, ignore
            ('_' | '\\', ..) => escaped = true,
            ('"', _, true, _) => is_quote = false,
            (_, _, true, _) => (),
            ('"', _, _, _) => is_quote = true,
            (';', ..) => return false, // first format only
            ('[', .., 0) => brackets += 1,
            ('[', ..) => brackets += 1,
            (']', .., 1) if hms => return true, // if closing
            (']', ..) => brackets = brackets.saturating_sub(1),
            ('d' | 'm' | 'h' | 'y' | 's' | 'D' | 'M' | 'H' | 'Y' | 'S', _, _, 0) => return true,
            _ => {
                if hms && s.eq_ignore_ascii_case(&prev) {
                    // ok ...
                } else {
                    hms = prev == '[' && matches!(s, 'm' | 'h' | 's' | 'M' | 'H' | 'S');
                }
            }
        }
        prev = s;
    }
    false
}
```

dimastbk:
Thanks for review! Sorry for regex :(

```
test bench_regex  ... bench:      11,279 ns/iter (+/- 186)
test bench_first  ... bench:       3,873 ns/iter (+/- 51)
test bench_second ... bench:       1,357 ns/iter (+/- 32)
```
I have run all proposals with `test_is_date_format` (with small fix for second - I have added detecting of string with A/P (AM/PM) only). Should I replace `is_custom_date_format` with code above? This implementation passes all tests from sheetjs, except one (same that original).
 
```rust
pub fn is_custom_date_format(format: &str) -> bool {
    let mut escaped = false;
    let mut is_quote = false;
    let mut brackets = 0u8;
    let mut prev = ' ';
    let mut hms = false;
    let mut ap = false;
    for s in format.chars() {
        match (s, escaped, is_quote, ap, brackets) {
            (_, true, ..) => escaped = false, // if escaped, ignore
            ('_' | '\\', ..) => escaped = true,
            ('"', _, true, _, _) => is_quote = false,
            (_, _, true, _, _) => (),
            ('"', _, _, _, _) => is_quote = true,
            (';', ..) => return false, // first format only
            ('[', ..) => brackets += 1,
            (']', .., 1) if hms => return true, // if closing
            (']', ..) => brackets = brackets.saturating_sub(1),
            ('a' | 'A', _, _, false, 0) => ap = true,
            ('p' | 'm' | '/' | 'P' | 'M', _, _, true, 0) => return true,
            ('d' | 'm' | 'h' | 'y' | 's' | 'D' | 'M' | 'H' | 'Y' | 'S', _, _, false, 0) => {
                return true
            }
            _ => {
                if hms && s.eq_ignore_ascii_case(&prev) {
                    // ok ...
                } else {
                    hms = prev == '[' && matches!(s, 'm' | 'h' | 's' | 'M' | 'H' | 'S');
                }
            }
        }
        prev = s;
    }
    false
}
```

tafia:
Yes please go ahead. Thanks a lot!

dimastbk:
I updated this pr.

> this module should be conditional to dates feature (#[cfg(feature = "dates")])
I think, it doesn't make sense with the new implementation.

tafia:
Thanks

tafia:
Thanks

dimastbk:
Before this PR:
```rust
Range {
    start: (0, 0),
    end: (3, 1),
    inner: [
        String("DateColWithBigInt"),
        String("StringCol"),
        Float(42441.0),
        String("Marc Johnson"),
        Float(42445.0),
        String("Jack Black"),
        Float(1e20),
        String("Timothy Brown"),
    ],
}
```

After this PR:
```rust
Range {
    start: (0, 0),
    end: (3, 1),
    inner: [
        String("DateColWithBigInt"),
        String("StringCol"),
        DateTime(42441.0),
        String("Marc Johnson"),
        DateTime(42445.0),
        String("Jack Black"),
        DateTime(1e20),
        String("Timothy Brown"),
    ],
}
```

Hm, maybe add the ability to extract float/string value from datetime/duration, like below?
```rust
    /// Try getting float value
    pub fn get_float(&self) -> Option<f64> {
        if let DataType::Float(v) | DataType::DateTime(v) = self {
            Some(*v)
        } else {
            None
        }
    }
...
    /// Try getting string value
    pub fn get_string(&self) -> Option<&str> {
        if let DataType::String(v) | DataType::DateTimeIso(v) | DataType::DurationIso(v) = self {
            Some(&**v)
        } else {
            None
        }
    }
```

fzumstein:
Right, thanks for pointing out the real change! in this case, I think all is good!

PR #306, opened 2023-05-05
Patrick dedication/add partial eq string slice to data type

patrick-dedication:
When writing test for our use-case we found that we can't do the following
```
let row = vec![DataType::String("value".to_string())];
assert_eq!(row, vec!["value"]);
```

### DIFF
--- a/src/datatype.rs
+++ b/src/datatype.rs
@@ -327,7 +327,7 @@ where
 }
 
 #[cfg(all(test, feature = "dates"))]
-mod tests {
+mod date_tests {
     use super::*;
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/datatype.rs
```diff
@@ -166,6 +166,15 @@ impl DataType {
     }
 }
 
+impl PartialEq<&str> for DataType {
```
tafia, line 169: Why not `T: AsRef<str>`? It'll be more useful I think

tafia (changes requested):

src/datatype.rs
patrick-dedication: Thanks for the hint, I tried it out.
If I got you right, this seems not to be legal, because they could impl AsRef<str> for i.e. f64, then we would have two conflicting implementations for PartialEq.
image

tafia (approved):

PR #308, opened 2023-05-12
ODS: handle table:number-rows-repeated attribute

dimastbk:
Hi. This PR adds support `table:number-rows-repeated` in `<table:table-row>` in ODS file, reference: http://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#__RefHeading__1418532_253892949. The test document was created in Microsoft Excel Online; Google Sheet also creates rows with `table:number-rows-repeated`, but only for empty rows.


tafia:
Thanks!

PR #277, opened 2022-10-20
251 date precision

jqnatividad:
Hi @tafia . Here's my attempt to resolve #251 originally reported by @usagi.

I added `once_cell` to get `excel_epoch` more efficiently, and I also bumped `chrono` to the latest release.

Issue #251, opened 2022-01-23
"dates" feature is not accurate enough, maybe


usagi:
The dates feature is not accurate enough.

Actual:

1. Make the xlsx file with the value `2021-10-15 19:00:00` of DateTime typed.
2. Read the value using `as_date()` with `features=["dates"]`
3. The result is "2021-10-15 18:59:59.999999761".

Expected:

- The result is "2021-10-15 19:00:00".

Guessing:

- Maybe, this constant is not accurate enough:
  - https://github.com/tafia/calamine/blob/master/src/datatype.rs#L127

FYI:

I was tested in the below code, and it was taken the correct result.
```
pub fn to_naive_from_excel_datetime(v: f64) -> NaiveDateTime
{
 let excel_epoch = NaiveDate::from_ymd(1899, 12, 30).and_hms(0, 0, 0);
 let h = v * 24f64;
 let m = h * 60f64;
 let s = m * 60f64;
 let ms = s * 1e+3f64;
 let excel_duration = Duration::milliseconds(ms as i64);
 excel_epoch + excel_duration
}
```

But, I know this implementation has an overheads and you don't like the explainly implements maybe. Therefore, I didn't PR and just write the issue only.

tafia:
I wouldn't mind a PR. In the worse case it'd be behind some feature gate.

PR #309, opened 2023-05-12
fix datetime precision, fix panic on date overflow

dimastbk:
Hi! This PR add two fixes for parsing excel datetime:

1. Additional fix for #251 and #277 - replace type coersion with `f64.round()`:
`0.25951736111111101 * 24f64 * 60f64 * 60f64 * 1e+3f64 = 22422299,99999999`
before: `22422299,99999999 as i64 = 22422299 = 06:13:42.299`
after: `22422299,99999999.round() as i64 = 22422300 = 06:13:42.300`
2. Fix panic on datetime overflow.


tafia:
Thanks!

PR #316, opened 2023-05-23
fixed #315: datetime parsing in xlsb, when only standard styles are present

### LINKED ISSUE #315: Date parsing with xlsb fails if there are no custom formats present
#296 by @ketbra introduced date support for `xlsb` but it fails to handle the built-in formats in case there are no custom-formatted cells present as in this case, the `BrtBeginFmts` section is missing (which is where the built-in date formats are currently being handled).

To reproduce: run `date_xlsb` test, but using the attached file instead of `date.xlsb`.

[date_builtin.zip](https://github.com/tafia/calamine/files/11543813/date_builtin.zip)

dimastbk:
This PR should close #315.

### DIFF
--- a/src/xlsb.rs
+++ b/src/xlsb.rs
@@ -109,7 +109,7 @@ impl std::error::Error for XlsbError {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 enum CellFormat {
     Other,
     Date,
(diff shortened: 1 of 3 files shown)


src/xlsb.rs
```diff
@@ -188,36 +188,48 @@ impl<RS: Read + Seek> Xlsb<RS> {
...
-                        Some(fmt) if is_custom_date_format(fmt) => CellFormat::Date,
-                        None if is_builtin_date_format_id(num_fmt.to_string().as_bytes()) => {
-                            CellFormat::Date
-                        }
-                        _ => CellFormat::Other,
-                    });
+                    for _ in 0..len {
+                        let _ = iter.next_skip_blocks(0x002C, &[], &mut buf)?; // BrtFmt
+                        let fmt_code = read_u16(&buf);
+                        let fmt_string = wide_str(&buf[2..], &mut 0)?.into_owned();
```
tafia, line 203:
```suggestion
                        let fmt_str = wide_str(&buf[2..], &mut 0)?;
                        let fmt = if is_custom_date_format(fmt_str) {
                            CellFormat::Date
                        } else {
                            CellFormat::Other
                        };
                        number_formats.insert(fmt_code, fmt);
```

To avoid allocating the fmt as we're only caring if it is a date or not (for the moment).

src/xlsb.rs
```diff
@@ -188,36 +188,48 @@ impl<RS: Read + Seek> Xlsb<RS> {
...
+                        number_formats.insert(fmt_code, fmt_string);
+                    }
+                }
+                0x0269 => {
+                    // BrtBeginCellXFs
+                    let _len = iter.fill_buffer(&mut buf)?;
+                    let len = read_usize(&buf);
+                    for _ in 0..len {
+                        let _ = iter.next_skip_blocks(0x002F, &[], &mut buf)?; // BrtXF
+                        xfs.push(read_u16(&buf[2..4]));
```
tafia, line 213: If I understand the [spec](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-xlsb/b0da927e-fefc-45b0-bdb9-38883737957f) correctly, `FMTS` always comes before `CELLXFS` so `number_formats` is guaranteed to be properly populated. As a result, we can update `self.format` directly and thus avoiding an extra `xls` vec.

```suggestion
                        xfs.push(read_u16(&buf[2..4]));
                        let fmt_code = read_u16(&buf[2..4];
                        
                        // if we assume we modify `is_builtin_date_format_id` fn
                        // adding in each match's branch the byte variant:
                        //  `b"14" => ... ` becomes `b"14" | &14u16.to_le_bytes()) => ...`
                        // we could then avoid allocating here too
                        
                        if is_builtin_date_format_id(&fmt_code.to_le_bytes()) {
                            self.formats.push(CellFormat::Date);
                        } else {
                            self.formats.push(number_formats.get(&fmt_code).copied().unwrap_or(CellFormat::Other));
                        }
```

tafia (approved):
Awesome, made a few suggestion (perf). Really not mandatory, let me know what you think.

src/xlsb.rs
```diff
@@ -188,36 +188,48 @@ impl<RS: Read + Seek> Xlsb<RS> {
...
+                    }
                 }
+                // BrtEndStyleSheet
+                0x0117 => break,
```
tafia, line 217: we could probably early exit when reaching [BrtEndCellXFs](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-xlsb/cccbca18-dcb2-4235-a123-827cf1ab11d2) ?

dimastbk: I think, we can exit after loop in BrtStartCellXFs, as you suggest below.

dimastbk:
Thanks for review! I have fixed all your suggestions.

tafia:
Thanks!

Very not important given that the impact should be minimal but I think we should maybe have `is_builtin_date_format_id` using `u16` instead of `&str` for its arguments, given how simple it is

```rs
fn is_date_code(code: u16) -> bool {
    matches!(code, 14..=22 | 45..=47)
}
```

PR #322, opened 2023-05-29
bump msrv to 1.59, add msrv to GitHub Action

dimastbk:
MSRV was added in #211. #312 and some other changes broken this MSRV. I think adding this test can help in the future. 

Also this PR bumps msrv to 1.59 (because zip requires 1.59), makes clippy happy (!matches macros) and fixes chrono deprecation warnings.


tafia:
Thanks!

PR #324, opened 2023-05-30
timedelta and 1904 datetime system

### LINKED ISSUE #318: Datetime support
Some small (in my opinion) bugs in datetime support, that I am aware of:
* support for duration in xls/xslx/xlsb (`[hh]:mm:ss` parsed as datetime):
* * `255:10:10.as_time() == None`
* * `255:10:10.as_datetime() == datetime(1900, 1, 9, 15, 10, 10)`
* fix duration in ods, when more than 24 hours (255:10:10 => PT255H10M10S.as_time() == None), see also https://github.com/chronotope/chrono/issues/579
* support for 1904 date system in xlsb/xlsx
* ~~improve some performance https://github.com/tafia/calamine/pull/316#issuecomment-1562195694~~

dimastbk:
This PR add support timedelta and 1904 datetime system in excel files (partially closed #318).

For 1904 datetime system I'm using a hack like in #299 and I'm not sure that it's good. Maybe something like?
```rust
pub enum DataType {
    /// Date or Time, bool - is_1904
    DateTime(f64, bool),
}
```

```
 name        old ns/iter  new ns/iter  diff ns/iter  diff %  speedup 
 bench_ods   74,811       75,899              1,088   1.45%   x 0.99 
 bench_xls   32,367       32,154               -213  -0.66%   x 1.01 
 bench_xlsb  139,909      141,788             1,879   1.34%   x 0.99 
 bench_xlsx  197,789      196,129            -1,660  -0.84%   x 1.01
```


tafia:
Thanks a lot one more time!

PR #328, opened 2023-06-03
fix incorrect date parsing caused by excel date bug

Geo-W:
Date earlier than `1900-03-01` will be parsed 1 day earlier, 
which is caused by excel incorrectly recognizing 1900 as leap year.

e.g.  `1900-01-05` => `Some(1900-01-04T00:00:00)`  

both `1900-02-28` and `1900-02-29` will be parsed as `Some(1900-02-28T00:00:00)` after fix.


tafia:
Thanks

PR #342, opened 2023-07-26
Fixup Range#get: add check for column out of range

A248:
I noticed that `Range::get` was lacking some attention, so I added the missing range check on the column index, as well as some documentation


tafia:
Thanks!

PR #345, opened 2023-07-31
docs: remove datetime notice from README

dimastbk:
Hi!

Yes, datetime support isn't perfect, but I think it's good enough for removing this comment from README.


tafia:
Agreed

Issue #343, opened 2023-07-27
Xls error: Unrecognized error: 0x3


7Towers:
When attempting to open a workbook, I'm getting the title error. 
```
Xls error: Unrecognized error: 0x3
```
<i>Edit</i>
Can someone provide me some tips to debug? ~~The sheet has some financial specifics, and customer names, so it's not easily shared. I may be able to create some dummy data, but would like to start with some tips to troubleshoot on my end, with the help of you pros~~. I was able to strip the sheet of all data, and I'm still having trouble opening it with this library. 
Sample sheet can be found [here](https://docs.google.com/spreadsheets/d/1Oc46NUkuZ_Uj5GW4bssMsFO0h3Wmn5E3/edit?usp=sharing&ouid=105796885453683062541&rtpof=true&sd=true)

The sheet is saved as an .xls. If I open the sheet in Excel and re-save as xlsx, then attempt to open the workbook in calamine, I still receive the same error. 

Nothing fancy on the opening code: 
```
    let mut workbook = match open_workbook_auto(path) {
        Ok(wb) => wb,
        Err(e) => {
            eprintln!("Error opening workbook: {}", e);
            return Err(e);
        }
    };
```
That code yields 
```
Error opening workbook: Xls error: Unrecognized error: 0x3
```

When I attempt to open these sheets using umya spreadsheets (another rust xls lib), I get a little descriptive error: 

```
ZipError: invalid Zip archive: Could not find central directory end
```
I am able to open the sheet in Excel, and with other frameworks like xlsx js (SheetJS), but that's a bit slower. 

There are many refs in this workbook to other workbooks in the same directory, as well as heavy use of macros and other content that I'm not familiar with, yet.

dimastbk:
Hi!

> Error opening workbook: Xls error: Unrecognized error: 0x3

It seems calamine can't handle "0x03 Blank string value." in [FormulaValue](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-xls/39a0757a-c7bb-4e85-b144-3e7837b059d7).

Could you attach a sample in GitHub? I can't download the original from Google Sheets.

7Towers:
Thanks @dimastbk. 

Here's the file
[Sample.xls](https://github.com/tafia/calamine/files/12182988/Sample.xls)

7Towers:
Update: 

[...]

PR #348, opened 2023-08-21
Allow handling empty string value in xls formula parsing

7Towers:
#343 
Simple fix to handle empty string values when parsing XLS formulas. Based on MS recommendations. Fixes the issue I've faced in my projects.


tafia:
Yes please add a test. Thanks!

dimastbk:
I think, `0x03` should return `DataType::String("".to_string())`, because empty cell != zero-lenght string cell. See for example [xlrd](https://github.com/python-excel/xlrd/blob/master/xlrd/sheet.py#L992) and [sheetjs](https://git.sheetjs.com/sheetjs/sheetjs/src/branch/master/bits/63_fbin.js#L50).

jqnatividad:
Hope you don't mind @7Towers , but I went ahead and implemented an alternate PR using @dimastbk 's recommendation to return an empty string, using the test file he provided.

Hopefully, when this gets merged, the next calamine release will be published by @tafia 😉 🤞

PR #349, opened 2023-08-23
Formula empty string value handling

jqnatividad:
alternative to @7Towers' PR (#348 ) but returning `DataType::String("".to_string())` instead of just `None` to align calamine with how xlrd and sheetjs handles it per @dimastbk .

Also includes test, using test file provided by @dimastbk


tafia:
Thanks!

PR #359, opened 2023-09-19
Support label cells for XLS

softdevca:


softdevca:
Labels store an XLUnicodeString  directly instead of an indexing into a string pool like LabelSst

tafia:
Thanks!!

PR #364, opened 2023-10-07
perf: pre-allocate buffers with `with_capacity()`

RoloEdits:
Pre-allocating removes a lot of regrows for the buffers.

## Benchmarks
Quick benchmark opening a 150k row, 17 column `.xlsx` file shows 15% increase in performance.

`master`:
```bash
868,545,530 ns/iter (+/- 7,365,438)
```
`perf/pre_alloc_buf`:
```bash
750,808,050 ns/iter (+/- 6,366,129)
```

## Profiles

Purple highlight is `alloc`.

Current `master`:
image

Change:
image

Changed from `vec![]` macro to `with_capacity()` as there was an introduction to a lot of instructions running.

`vec![0;N]`:
image

`with_capacity(N)`:
image

### DIFF
--- a/src/xlsx.rs
+++ b/src/xlsx.rs
@@ -1222,8 +1222,8 @@ fn read_string(
     xml: &mut XlsReader<'_>,
     QName(closing): QName,
 ) -> Result<Option<String>, XlsxError> {
-    let mut buf = Vec::new();
-    let mut val_buf = Vec::new();
+    let mut buf = Vec::with_capacity(1024);
+    let mut val_buf = Vec::with_capacity(1024);
     let mut rich_buffer: Option<String> = None;
     let mut is_phonetic_text = false;
     loop {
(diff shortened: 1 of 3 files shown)


src/xlsx.rs
```diff
@@ -1222,8 +1222,8 @@ fn read_string(
     xml: &mut XlsReader<'_>,
     QName(closing): QName,
 ) -> Result<Option<String>, XlsxError> {
-    let mut buf = Vec::new();
-    let mut val_buf = Vec::new();
+    let mut buf = Vec::with_capacity(1024);
+    let mut val_buf = Vec::with_capacity(1024);
```
RoloEdits, line 1226: `val_buf` in particular had the biggest impact when profiling with this change

RoloEdits:
Marginal improvement but still an improvement:

pre `2881d13`:
image

post:
image

tafia:
Awesome, thanks!

PR #369, opened 2023-10-25
use nested crate to store xlsx shared strings

tafia:
Use `nested::Nested` to store xlsx shared strings.

My tests show a marginal improvement in terms of speed but memory impact might be noticeable for pathologically large SharedString files.


jmcnamara:
I am going to close this to clean up the backlog. If someone wants to revive this approach they can open a new pull request.

Issue #362, opened 2023-10-04
Small String Optimization


RoloEdits:
Considering the domain, as well as the specification, it seems a lot of gains could be made from using something like [smartstring](https://lib.rs/crates/smartstring) in-place of `String` for some performance gains. 

It could also replace some of the `Vec<u8>` as keys in `BTreeMap` as the in-lined string has better cache locality. And even if it does need to get allocated then its not any worse then a `Vec`.

tafia:
I am not certain of the impact it could have but indeed strings are mostly small and immutable and it seems like a good fit.

RoloEdits:
I tried to make a quick and dirty patch just to see if there would be any immediate differences.

Here are the flamegraphs. Purple is `alloc`.

Current `std::string::String`
![image](https://github.com/tafia/calamine/assets/12489689/1da3362c-13d7-47fd-829d-41d9594c3142)

SSO with `smartstring::String`
![image](https://github.com/tafia/calamine/assets/12489689/6f82b7a4-04ef-4ca8-99f3-5c519278dca8)

The biggest "win" is this block disappearing
![image](https://github.com/tafia/calamine/assets/12489689/0ebe6c9a-0cce-4370-ac5e-c51550cc5b8b)

I think it could be worth more time to do a proper implementation. I'll try to get a proper PR for this when I have some time.

tafia:
Thanks a lot for looking at it!

RoloEdits:
With the current way the library is designed, which directly exposes the `String` in the cell value, as well as the other places a `String` is exposed, hiding the new type is impossible without changes. And if  `&str` are stored in the cells, then large changes would be needed.

I can try to hack together a feature flag for it that would expose the new SSO string directly, and use it where possible, but this is going to lead to a bit of a code mess and also further lock us into the SSO implementation we go with, as the new type would be part of the public API, so this cant even be an implementation detail.

How would you feel if I write up a proposal about working towards a version 1.0 with some changes to accommodate better ergonomics and better encapsulation? I have looked through some of the issues and can try to address them as I go, as well as at least start some discussion on how writing spreadsheets would go. With the goal, ultimately, to make `calamine` the goto for spreadsheet related matters in rust.

tafia:
I'd be happy to see how it could go yes but i'm afraid it may take quite a lot of your time.
I don't mind breaking the public api too much if necessary, this is still a pre 1.0. 

[...]

PR #370, opened 2023-10-25
support DataTypeRef for shared strings and worksheet CellsReader

tafia:
Here is another iteration at trying to reduce memory usage.

Introduce a new `DataTypeRef` which is the same as `DataType` but with an additional `SharedString` variant.

The test case from #362 should be amended with:
```rs
    // Get worksheet
    let sheet = excel
        .worksheet_range_ref("NYC_311_SR_2010-2020-sample-1M")
        .unwrap()
        .unwrap();
```

NOTE:
Given that large test file, we may revisit a protection used against files with mostly blank cells:
https://github.com/tafia/calamine/pull/370/files#diff-bedb7172be56842bcb4be66655e6eb625dadc8a029fe3c87c6657768ef1fc8b9R662

### DIFF
--- a/benches/basic.rs
+++ b/benches/basic.rs
@@ -17,7 +17,6 @@ fn count<R: Reader<BufReader<File>>>(path: &str) -> usize {
         count += excel
             .worksheet_range(&s)
             .unwrap()
-            .unwrap()
             .rows()
             .flat_map(|r| r.iter())
             .count();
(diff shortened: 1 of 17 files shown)


tafia:
It remains to be seen if this can be ported to other file formats.

tafia:
cells reader and `DataTypeRef` have been ported to xlsb. Ods and Xls may not be possible as we need to read the whole file at once.

PR #375, opened 2023-11-23
Fixes #374

### LINKED ISSUE #374: Text is truncated for xls document with BIFF5
I came across the file which, when read using calamine, had all text labels truncated (first and a last character of all text cells were missing), same issue with sheet tab names. It looks like the current implementation only works for newer versions of Microsoft Excel. 

I'm attaching the example file: [biff5_write.xls](https://github.com/tafia/calamine/files/13445320/biff5_write.xls)

saks:
This PR fixed #374. The logic is inspired by the implementation of https://github.com/SheetJS/sheetjs/. Example file is copied from https://github.com/SheetJS/test_files.

I do realize that this change fixes an issue that can only occur when file was created using really old software but customers still upload such files, hence I truly believe it is a useful addition to this project.

Please let me know if there is anything that I can improve/fix.


src/xls.rs
```diff
@@ -455,10 +461,60 @@ impl<RS: Read + Seek> Xls<RS> {
...
+        // BIFF5
+        0x0500 => 5,
+        // BIFF4
+        0x0400 => 4,
+        // BIFF3
+        0x0300 => 3,
+        // BIFF2
+        0x0200 | 0x0002 | 0x0007 => 2,
+        _ => {
+            if r.data[4..].len() > 6 {
```
tafia, line 491: This should probably be checked for all Bofs? Spec says that is `r.data.len() == 16`?

src/xls.rs
```diff
@@ -455,10 +461,60 @@ impl<RS: Read + Seek> Xls<RS> {
...
+                    expected: 6,
+                    found: r.data.len(),
+                });
+            } else {
+                8
+            }
+        }
+    };
+
+    if biff_version == 0 && dt == 0x1000 {
```
tafia, line 503: 1. How come a biff_version = 0 is valid? Even looking at spces from 2012 (the oldest one available on Microsoft website) I don't see it mentioned.
2. I think I'd prefer having these checks in the `_` branch only
3. I am not sure where this logic is coming from? I didn't find it in sheet.js?

tafia (commented):
Thanks a lot!
I have made a few comments, do you mind using an enum and specifying in comments where the logic for biff_version == 0 is coming from? I can't find it. Thanks!

src/xls.rs
saks: 1. It doesn't appear that the BIFF0 is a valid version to me either. However it looks like this logic is related to how the beginning of BOF record is interpreted.
2. Agree! 
3. [This](https://github.com/SheetJS/sheetjs/blob/5b4806bff01a9f4289b4c04e8a60a6c4c52fe93a/xlsx.js#L19383) is where I found it. Not sure all of that logic is in agreement with the rest of the code I've seen in the same file though. I guess some of it is not making sense to me now. Let me clean it up a little.

src/xls.rs
saks: Agree, although I'm not sure if we need this check at all. Initially I've copied it from sheetjs, but now I don't see a purpose to be honest. Especially if I cannot write a test (I don't have an example xls file) to hit this line. What do you think?

saks:
> Thanks a lot! I have made a few comments, do you mind using an enum and specifying in comments where the logic for biff_version == 0 is coming from? I can't find it. Thanks!

Thanks for the review! PR updated. Please let me know if changes are making sense.

src/xls.rs
tafia: Thanks, 
As we do not have a proper test file for this and I don't understand why it is here, could we perhaps comment this part of the code, so if someone has an issue we can uncomment and put a proper test, in the meantime staying "true" to the spec seems like a better choice.

tafia (approved):

src/xls.rs
```diff
@@ -464,57 +464,53 @@ impl<RS: Read + Seek> Xls<RS> {
...
+                Biff::Biff8
             }
         }
+        _ => Biff::Biff8,
     };
 
-    if biff_version == 0 && dt == 0x1000 {
-        biff = 5;
-    }
-    if biff == 8 && biff_version == 0 && dt == 16 {
```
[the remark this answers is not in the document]
saks: It looks like this logic was used for a weird file (https://github.com/SheetJS/sheetjs/issues/1220) and I'm not sure we need/want to support this since I cannot really find a test file for it. I'm still trying to understand the reason for `dt == 0x1000`.

src/xls.rs
```diff
@@ -455,10 +461,57 @@ impl<RS: Read + Seek> Xls<RS> {
...
+    };
+
+    let biff = match biff_version {
+        0x0200 | 0x0002 | 0x0007 => Biff::Biff2,
+        0x0300 => Biff::Biff3,
+        0x0400 => Biff::Biff4,
+        0x0500 => Biff::Biff5,
+        0x0600 => Biff::Biff8,
+        0 => {
+            if dt == 0x1000 {
```
saks, line 498: I wasn't able to find an example of `biff5` file that would trigger this logic. Please let me know if you'd like me to remove this code.

src/xls.rs
```diff
@@ -464,57 +464,53 @@ impl<RS: Read + Seek> Xls<RS> {
...
-
     Ok(Bof { biff })
 }
 
 /// BoundSheet8 [MS-XLS 2.4.28]
 fn parse_sheet_metadata(
     r: &mut Record<'_>,
     encoding: &XlsEncoding,
-    biff: u8,
+    biff: &Biff,
```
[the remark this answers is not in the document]
saks: done

src/xls.rs
saks: Just to clarify: I'm talking about removing this code since, if we comment it out, as you've suggested earlier, we'd have no use for `dt` variable. Then I guess we'll need to comment out more code. I agree that following the spec is the best option, especially with no evidence that we'll ever encounter files that can trigger this logic.

PR #376, opened 2023-11-28
Fixes #102

### LINKED ISSUE #102: XLSX with passwords
It seems that no Rust crate at the moment supports encrypted spreadsheets but is there any plans? It seems that password support is by no means a trivial feature, since there are many different schemes to encrypt an XLSX file. I don't even know if some are more common than others: http://poi.apache.org/encryption.html

I think that the Apache POI library is the gold standard standard for supporting stuff: https://github.com/apache/poi

As one step towards a better direction, it would be nice if it could recognize and report the error correctly, as it does with the XLS format.

saks:
This PR attempts to help with the detection of password protected XLSX files (#102). I guess there are many improvements that can be made to this first naive attempt, but I'd like to get the conversation started.

Even in this basic form it already solves the problem we have on our project (we won't have to call `file` as a sub-process).

Please let me know what can be improved, happy to get any feedback.

Thanks!


tafia:
Thanks!

PR #379, opened 2023-12-03
 Use `chunks` instead of `chunks_exact`

jqnatividad:
The `chunks_exact` method returns chunks of exactly 4 bytes, and any leftover bytes at the end of the slice are ignored.

Since we know that the input slice will always be a multiple of 4 bytes per the assert, we can use the `chunks` method instead, which is slightly faster because it doesn't need to redundantly check for leftover bytes.


tafia:
Thanks!

PR #386, opened 2023-12-18
Detect password protected xls files

### LINKED ISSUE #385: Password protected XLS is not detected
We have lots of password protected XLS files that cannot be open with calamine currently. The error we get is always consistent:
```
Cfb(CodePageNotFound(7766))
```
The interesting part is that the value that `CodePageNotFound` wraps is different for different files. This value never looks like a valid `codepage` though.

Attached example file is an empty password protected spreadsheet (saved as "**Excel97**" compatible) allows to reproduce this issue.

[pass_protected.xls](https://github.com/tafia/calamine/files/13709218/pass_protected.xls)

saks:
This PR fixes #385. It looks like [Protect](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-xls/b32708d5-35d2-4005-9488-abbfc319ae1f) record that is currently used for password protected files detection may not be the best option. There are two reasons:
1. It is easy to find/create a file that triggers unrelated error.
2. [Spec](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-xls/24fed501-1f2e-435b-a371-ad29a57dba58#gt_edd09e04-ab93-4769-ae34-2cad606fa829) suggests that:
>  A property that is applied manually to a file or a portion of a file, with or without a password, and that helps prevent users from accidentally or deliberately changing, moving, or deleting data.

this property may signify another kind of protection but not necessarily the password one.
I've noticed that some libraries are trying to detect "[FilePass](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-xls/cf9ae8d5-4e8c-40a2-95f1-3b31f16b5529)" instead and it works better for problematic files that we have.

I was not able to determine if it is necessary to keep logic related to "Protect" record in place (my changes do not break any tests), please suggest if we need to keep it.


tafia:
Thanks!

Issue #339, opened 2023-07-02
Feature Request: proposal for new types addition and optional integration with `rust_xlsxwriter`


lucatrv:
I would like to open a discussion to add a few convenience data types with [`serde::Deserialize`](https://docs.rs/serde/latest/serde/trait.Deserialize.html) trait implementations. With these new data types it would become trivial to deserialize Excel cells which normally contain a certain type (for instance a float) but may also contain other values such as strings, errors, etc. This proposal also discusses the optional integration with the stellar [`rust_xlsxwriter`](https://crates.io/crates/rust_xlsxwriter) crate, and the opportunity to use its Excel specific [`rust_xlsxwriter::ExcelDateTime`](https://docs.rs/rust_xlsxwriter/latest/rust_xlsxwriter/struct.ExcelDateTime.html) type to represent date and/or time instead of the generic [`chrono::naive::NaiveDate`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html).

I am splitting this proposal into separate sections within a single Feature Request because all points are related:

1. Rename `DataType::as_f64` into `DataType::as_float` and `DataType::as_i64` to `DataType::as_int`. This is for consistency with `DataType::is_float` / `DataType::get_float` and `DataType::is_int` / `DataType::get_int`. and with the proposed new data types (see point 2). This is not a loss of generality because numerical values in Excel are anyway stored as `f64` types, see reference [here](https://docs.rs/rust_xlsxwriter/latest/rust_xlsxwriter/struct.Worksheet.html#method.write_number). This change could be introduced by first adding `DataType::as_float` and `DataType::as_int` and moving `DataType::as_f64` and `DataType::as_i64` to undocumented / deprecated state, before final removal.

2. For each `DataType` variant, add a corresponding `VariantOrString` type, and implement the `Default`, `PartialEq`, `Eq` when appropriate for the type (see [`std::cmp::Eq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html)), and [`serde::Deserialize`](https://docs.rs/serde/latest/serde/trait.Deserialize.html) traits (plus the optional trait [`rust_xlsxwriter::IntoExcelData`](https://docs.rs/rust_xlsxwriter/latest/rust_xlsxwriter/trait.IntoExcelData.html) when the feature `rust_xlsxwriter` is enabled, see point 3).
[...]

PR #389, opened 2023-12-19
Rename `DataType` methods `as_f64` `as_i64` into `as_float` `as_int`

lucatrv:


lucatrv:
@tafia can this be merged or is anything else needed?

tafia:
There are conflicts, sorry it is because of another PR I just merged.

lucatrv:
@tafia, sorry I just had a second thought about this, maybe the reason for the current nonuniform naming is because:

- `is_float` means: check if the `Data` variant is `Float`.
- `get_float` means: if the `Data` variant is `Float`, get its content.
- `as_f64` means: try to convert any `Data` variant into an `f64` type.

In this case, maybe the current naming is correct. Please let me know your thought, then I will either close this PR or resolve its conflicts for merging.
I am ready to create a new PR to add deserialize helping function which will work similarly to [`csv::invalid_option`](https://docs.rs/csv/latest/csv/tutorial/index.html#handling-invalid-data-with-serde), but I am stuck due to this PR, so I hope you'll let me know soon.
Thanks

tafia:
Sorry for the late answer. Yes this was the idea but I too am unsure about the names.
If you are unsure, then let's not do anything so it won't break too much code.
I let it opened for now for other people to comment either way.

lucatrv:
> Sorry for the late answer. Yes this was the idea but I too am unsure about the names. If you are unsure, then let's not do anything so it won't break too much code. I let it opened for now for other people to comment either way.

Yes I agree, so I will create a PR to close issue #339 using `as_f64` and `as_i64`.

lucatrv:
I am closing this because I think that the current naming is fine

PR #394, opened 2024-01-13
refactoring DateTime(f64) to DateTime(ExcelDateTime)

dimastbk:
Hi!

This PR introduces storing format of cell and is_1904 flag in datetime struct for correcting conversation to any format.

It's a partial fix for #339 and this [hack](https://github.com/tafia/calamine/issues/339#issuecomment-1864807355).

cc @lucatrv


src/datatype.rs
```diff
@@ -181,7 +180,7 @@ impl DataType {
                 let secs = days * 86400;
                 chrono::NaiveDateTime::from_timestamp_opt(secs, 0)
             }
-            DataType::Float(f) | DataType::DateTime(f) => {
+            DataType::Float(f) => {
```
dimastbk, line 183: I think, it (Float/Int conversation) isn't necessary and can be removed in the future.

src/datatype.rs
```diff
@@ -407,6 +403,73 @@ impl<'a> From<DataTypeRef<'a>> for DataType {
...
+        Some(chrono::Duration::milliseconds(ms.round() as i64))
+    }
+
+    /// Try converting data type into a datetime
+    #[cfg(feature = "dates")]
+    pub fn as_datetime(&self) -> Option<chrono::NaiveDateTime> {
+        let excel_epoch = EXCEL_EPOCH.get_or_init(|| {
+            chrono::NaiveDate::from_ymd_opt(1899, 12, 30)
+                .unwrap()
+                .and_hms_opt(0, 0, 0)
```
tafia, line 452: I know this is just a copy paste, but `and_time(NaiveTime::MIN)` doesn't require `unwrap`. Not sure if it is more performant (cannot fail) or it is optimized out, but seems better?

src/datatype.rs
tafia: Not sure we can discard it. I wouldn't trust the excel file to have it properly encoded every time and I'd expect people to try getting a datetime anyway.
On the other maybe it is better to create an `ExcelDateTime` and call `as_datetime`?

tafia (approved):
Thanks!

src/datatype.rs
dimastbk: > On the other maybe it is better to create an ExcelDateTime and call as_datetime?

Thanks, I dit it.

tafia:
I'm sorry the the last PR added conflict to yours (`DataType` has been renamed to `Data`).

dimastbk:
@tafia fixed

dimastbk:
We can de duplicate some code with this - https://github.com/dimastbk/calamine/pull/1/files, but it seems less readable.

dimastbk:
Thanks, I prefer second variant.

tafia:
Thanks!

PR #397, opened 2024-01-19
derive Clone on iterators

Expurple:
I need this feature in a dependent project and there seem to be no obstacles or hard-to-keep semver guarantees. So I went ahead and added it. Let me know if this can't be merged without tests, although there's no hand-written logic here


tafia:
Thanks!

PR #399, opened 2024-01-31
feat: support bools in DataType.as_f64 and DataType.as_i64 variants

lukapeschke:
Hello @tafia , I figured it would be nice to have an implicit conversion from bool to i64/f64 when using `as_{i64,f64}` , do you believe this makes sense as well ?

I haven't modified `as_string` yet because I thought it'd maybe be unexpected behaviour, but I'd gladly add if as well if you believe it makes sense :slightly_smiling_face:


tafia:
Thanks, I think it is reasonable yes.

PR #405, opened 2024-02-10
Support to get merged region

gy0801151351:
For the Xlsx files, we have a requirement. We need to determine the region of a merged cell. And based on the row range or column range of this range, find other rows or other cells listed in this range. Therefore, we create this PR to get the region of the merged cells.

For example:
The range of the merged cells is A1 to A2, the position of cell 2 is B1, the position of cell 3 is B2, and the position of cell 3 is B3. We need to read the data from B1 and B2 according to the range of this merged cell, because B1 and B2 are within the row range of the merged cell. Similarly, if the merged cells range is from A1 to A3, we need to read the data from B1, B2, and B3.


tafia:
Thanks!

jmcnamara:
@gy0801151351 

Did you have a reason, at the time, to return the file path along the the sheet name and range from the function calls:

```
("Sheet2", "xl/worksheets/sheet2.xml", Dimensions { start: (0, 0), end: (3, 0) })
```

It doesn't seem like useful information to most users. The `sheet_name` and `Dimension` should be enough and in the case of `merged_regions_by_sheet()` the `Dimensions` should be enough.

Also, is this block actually required: https://github.com/tafia/calamine/blob/master/src/xlsx/mod.rs#L499-L500

The code compiles and the tests pass without it.

PR #409, opened 2024-02-20
feat: added is_error and get_error methods for the DataType trait

lukapeschke:
The `DataType` trait does not allow to access cell errors yet, this adds support for it


lukapeschke:
@tafia  could you please have a look ? :pray:

PR #411, opened 2024-02-27
feat: use std::cell::OnceCell instead of the once_cell crate

softdevca:
Use `std::cell::OnceCell` instead of the `once_cell` crate to reduce dependencies.  Requires bumping MSRV to Rust 1.70.


softdevca:
Closing until actions can be updated.

PR #412, opened 2024-02-27
feat: Use std::cell::OnceLock instead of the once_cell crate

softdevca:
Use `std::cell::OnceLock` instead of the `once_cell` crate for fewer dependencies.  Requires bumping MSRV to Rust 1.71.


tafia:
Thanks

PR #413, opened 2024-03-01
feat: add `Range::headers` method

### LINKED ISSUE #327: Please add `Range::headers() -> Vec<String>`
It is often required to get the first row values of a range deserialized to strings (which are the range headers), so IMHO it would be useful to add a specific helping method `headers` to `Range`, which would be equivalent to calling `rows().next()` and then converting to `String`. The [`csv` crate](https://crates.io/crates/csv) has a similar method, see [tutoral](https://docs.rs/csv/latest/csv/tutorial/index.html#reading-headers) and [documentation](https://docs.rs/csv/latest/csv/struct.Reader.html#method.headers).
To convert each `DataType` value to `String`, the `headers` method should call `ToString::to_string`.

lucatrv:
Closes #327

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -734,6 +734,34 @@ impl<T: CellType> Range<T> {
     }
 }
 
+impl<T: CellType + fmt::Display> Range<T> {
+    /// Get range headers.
+    ///
+    /// # Examples
+    /// ```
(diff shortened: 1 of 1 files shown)


tafia:
Thanks

Issue #391, opened 2024-01-03
Formula detection not working for dragged formulas


yzobus:
There is seemingly still a problem in detecting formulas when they are created via dragging another formula.

For example, I set A1 to 1 and define a formula that is "=A1 + 1" in the field A2 and drag it down to field A5. 
In the xlsx file on can see that in all fields from A2:A5, a formula is defined, however, only the ones in A2 and A3 are detected.

The reason for this seems to lie in the formatting of the xml file, whose sheetdata will look like this:
<sheetData>
<row r="1" spans="1:1" x14ac:dyDescent="0.3"><c r="A1"><v>1</v></c></row>
<row r="2" spans="1:1" x14ac:dyDescent="0.3"><c r="A2"><f>A1+1</f><v>2</v></c></row>
<row r="3" spans="1:1" x14ac:dyDescent="0.3"><c r="A3"><f t="shared" ref="A3:A5" si="0">A2+1</f><v>3</v></c></row>
<row r="4" spans="1:1" x14ac:dyDescent="0.3"><c r="A4"><f t="shared" si="0"/><v>4</v></c></row>
<row r="5" spans="1:1" x14ac:dyDescent="0.3"><c r="A5"><f t="shared" si="0"/><v>5</v></c></row>
</sheetData>

From this view, the output of the worksheet_formula functions makes sense, because there is no value stored between <f> and </f>, but a defined shard functionality. 

I think it would be nice to also detect formulas which are created by dragging, however, it also seems to be a nightmare to correctly get the formulas

tafia:
Can you share a workbook?

ling7334:
# shared formula issue
`f` tag with no specify value, can be calculated with reference to the shared formula.

consider a worksheet xml below
```xml

<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"
    xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" mc:Ignorable="x14ac xr xr2 xr3"
    xmlns:x14ac="http://schemas.microsoft.com/office/spreadsheetml/2009/9/ac"
    xmlns:xr="http://schemas.microsoft.com/office/spreadsheetml/2014/revision"
    xmlns:xr2="http://schemas.microsoft.com/office/spreadsheetml/2015/revision2"
[...]

PR #418, opened 2024-04-02
feat: add shared formula logic

ling7334:
As #391 mentioned, `f` tag can use shared formula which doesn't has real value.
formula need to be calculated, and offset can either be in row or column.

In this PR, `load_shared_formula` doesn't applied, because it take a full scan of file.


src/xlsx/cells_reader.rs
```diff
@@ -165,8 +187,136 @@ impl<'a> XlsxCellReader<'a> {
...
+                                                while self.formulas.len() < shared_index as usize {
+                                                    self.formulas.push(None);
+                                                }
+                                                self.formulas.push(Some((f, offset_map)));
+                                            }
+                                        }
+                                        if shared_ref.is_none() {
+                                            // shared formula
+                                            let cell_regex =
+                                                Regex::new(r"\b[A-Z]{1,3}\d+\b").unwrap();
```
tafia, line 275: Do not build it in a loop.

ling7334:
close this PR for trying anoter concept.

PR #423, opened 2024-04-10
Replace `zip` with `zip_next`

Pr0methean:
The `zip` crate hasn't been updated in almost a year, and has a number of bugs that can cause panics when trying to process an invalid zip file. This PR replaces it with `zip_next`, a fork I actively maintain.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ encoding_rs = "0.8"
 log = "0.4"
 serde = "1.0"
 quick-xml = { version = "0.31", features = ["encoding"] }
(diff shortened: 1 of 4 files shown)


tafia:
Did you notice any performance impact?

Pr0methean:
I haven't run the benchmarks, but I can assure you that a regression is highly unlikely.

Pr0methean:
Smart move. I was invited to take over the `zip` name so that the upgrade path would be more visible.

Pr0methean:
They already have; `zip-next` is now a wrapper around `zip`, published from
the same repository as `zip-next` used to be published from.

PR #425, opened 2024-04-16
Shared formula

### LINKED ISSUE #391: Formula detection not working for dragged formulas

ling7334:
Resolve #391
reference non-existence value result in None, hence reference incoming formula is not possible without a full scan of file.


tafia:
Thanks!

PR #432, opened 2024-05-08
Feat: check for memory errors on reading xlsx files and proper formatting of csv sample code

SgtPepper47:
Updates to the Excel range creation to check the machine memory against the memory required for the range. 
Updates to the sample Excel to csv writer to identify the memory error and to write out a properly formatted csv file with commas instead of semicolons.

### DIFF
--- a/examples/excel_to_csv.rs
+++ b/examples/excel_to_csv.rs
@@ -24,9 +24,13 @@ fn main() {
     let dest = sce.with_extension("csv");
     let mut dest = BufWriter::new(File::create(dest).unwrap());
     let mut xl = open_workbook_auto(&sce).unwrap();
-    let range = xl.worksheet_range(&sheet).unwrap();
-
-    write_range(&mut dest, &range).unwrap();
+    let range = xl.worksheet_range(&sheet);
+    if range.is_err() {
+        /// Capture error however you want
+        std::process::exit(1);
+    } else {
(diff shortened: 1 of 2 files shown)


tafia:
Could you add a test showing how this can fail?

SgtPepper47:
tafia,
Here is an example of a file with a sheet that fails to load into memory. Essentially, it will depend on your machine, but an Excel sheet has a finite range of about 1 million rows to 16k columns.  I just run the example script to convert 'Sheet1' to a csv file.
[Large_Sheet.xlsx](https://github.com/tafia/calamine/files/15313633/Large_Sheet.xlsx)
Thanks.

tafia:
tests are failing.

dimastbk:
May be we should check usize::MAX in loop, when add cell to `cells`? And this should be backported to other formats too.

dimastbk:
@SgtPepper47, sorry, cells.len() == usize::MAX

tafia (commented):
I am not sure I understand why you would reserve some capacity on `cells`. If anything you actually want to `shrink_to_fit`. The capacity and the checks should be reserved in the `from_sparse` function instead.

An alternative would also be to enhance the `Range` to be an enum, either `sparse` or `dense`. Basically there should be some checks to determine when using a sparse representation is better (like here, very large dimension but very few cells). 

This is of course much more work but it would be better.

Issue #440, opened 2024-06-04
Reading rich text formatted cells


FlixCoder:
Thank you for the great library! :)

Somewhat related to #424, #404 #427, but after all quite different:
I want to read Excel sheets with cells that have individually formatted values, but not on the whole cell, but only part of the text. So it is not the cell style, but the rich text value inside the cell that interests me.

Example in shared strings (cell value displayed as: "1, <ins>1</ins>, 2"):
```xml
        <si>
		<r>
			<t xml:space="preserve">1, </t>
		</r>
		<r>
			<rPr>
				<u />
				<sz val="8" />
				<rFont val="Arial" />
				<family val="2" />
			</rPr>
			<t>1</t>
		</r>
		<r>
			<rPr>
				<sz val="8" />
				<rFont val="Arial" />
				<family val="2" />
			</rPr>
			<t>, 2</t>
		</r>
	</si>
```

I am in the unfortunate situation that the formatting decides the meaning of the content 😅 

Previous art / other implementations: https://docs.rs/edit-xlsx/0.4.4/edit_xlsx/struct.RichText.html : https://docs.rs/edit-xlsx/0.4.4/src/edit_xlsx/api/cell/rich_text.rs.html#5-7 as well as https://docs.rs/umya-spreadsheet/latest/umya_spreadsheet/structs/struct.RichText.html

Depending on the scope and agreement on the implementation, I might be interested in contributing this soon.
I suppose it would need another data type of cell values and a similar RichText implementation? Not sure how parsing exactly works and how it looks like in ODS..

tafia:
Thanks for opening the issue and stating it as clearly. I'll be more than happy to review any PR :)

jmcnamara:
Closing in favour of #538.

PR #441, opened 2024-06-06
feat: Implement RichText support for XLSX

FlixCoder:
For #440 .

I implemented a RichText representation and changed the `Data` representation to use it. The parsing is so far only implemented for XLSX, the other formats continue to only ever produce `Data::String` even if it is actually rich text content.

The shared strings also have to be rich text now ^^

I think it turned out quite well, let me know what you think :)


src/de.rs
```diff
@@ -657,11 +658,12 @@ impl<'a, 'de> serde::Deserializer<'de> for DataDeserializer<'a> {
     {
         match self.data_type {
             Data::String(v) => visitor.visit_str(v),
+            Data::RichText(v) => visitor.visit_str(v.text()),
             Data::Empty => visitor.visit_str(""),
-            Data::Float(v) => visitor.visit_str(&v.to_string()),
-            Data::Int(v) => visitor.visit_str(&v.to_string()),
-            Data::Bool(v) => visitor.visit_str(&v.to_string()),
-            Data::DateTime(v) => visitor.visit_str(&v.to_string()),
+            Data::Float(v) => visitor.visit_string(v.to_string()),
```
tafia, line 663: Why did this change?

src/style.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Is this just plain text?
+    #[inline]
+    pub fn is_plain(&self) -> bool {
+        self.formatted.iter().all(|(_len, f)| f.is_default())
+    }
+
+    /// Get the full text value without formatting.
+    #[inline]
+    pub fn text(&self) -> &String {
```
tafia, line 43: `&str` ?

src/style.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Deconstruct this into a plain string with the text.
+    #[inline]
+    pub fn into_text(self) -> String {
+        self.text
+    }
+
+    /// Add a formatted text element to the rich text.
+    #[inline]
+    pub fn add_element(&mut self, element: RichTextPart<'_>) -> &mut Self {
```
tafia, line 55: Not very important but I'd expect `add_xxx`, `push_xxx` and `append_xxx` to return an element that would be evicted or an index. 
Here this is more like a builder pattern so a `with_element` seems better.

src/style.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Bold?
+    pub bold: bool,
+    /// Italic?
+    pub italic: bool,
+    /// Underlined?
+    pub underlined: bool,
+    /// Striked?
+    pub striked: bool,
+    /// Font size.
+    pub size: f64,
```
tafia, line 107: Do we really need that precision?

src/style.rs
```diff
@@ -0,0 +1,162 @@
...
+pub enum Color {
+    /// Default, no color.
+    #[default]
+    Default,
+    /// Using the [colorIndex property](https://learn.microsoft.com/en-us/previous-versions/office/developer/office-2007/cc296089(v=office.12))
+    /// `index_id`: color index id.
+    Index(u8),
+    /// Use the Theme color format, i.e. leave the color decision to the current theme of xlsx.
+    /// You can determine the theme color id and tint by `theme_id` and `tint`.
+    Theme(u8, f64),
```
tafia, line 159: same comment, do we need that precision?

src/datatype.rs
```diff
@@ -339,9 +344,9 @@ pub enum DataRef<'a> {
     /// Float
     Float(f64),
     /// String
-    String(String),
+    String(RichText),
```
tafia, line 347: I'm a bit concerned about the overhead of keeping a `RichText` even if this is not required. Can't we have a `RichText` variant here too?

tafia (commented):
Awesome, this looks really neat!
I made a few tiny comments, the most important one being of not having a `DataRef::RichText` variant.

src/de.rs
FlixCoder: Entirely unrelated, sorry ^^ I suppose serde will need one allocation less for each of those if you pass the string you already have

src/style.rs
FlixCoder: In general, it is better to return the type it is, but accept a str or slice in parameters, as that allows the greatest freedom for the user.
Returning a &str restricts the user and does not allow to call methods from String (not sure if there are any interesting), but the user can easily get a &str. The other way around, it is not possible to get a &String from &str without an additional allocation.

But of course it might be annoying to have to convert to &str, so I have no hard feelings, which one should I pick?

src/style.rs
FlixCoder: I think I can easily return the index and have it be a push, as it really behaves more like a push.

src/style.rs
FlixCoder: Same "I don't really know" :D I'll use a f32 and we can see if somebody ever complains ^^

src/datatype.rs
FlixCoder: Hmm, I'll have to check again. I had some problems when I tried, but I don't remember what and it might be resolved now? :D I'll come back when I tried

src/style.rs
tafia: If you think `u16` is enough, then probably better to use it. I don't really care what other crates are doing. We'll revisit later if needed. Thanks!

src/datatype.rs
tafia: Awesome!

src/style.rs
FlixCoder: Renamed to `push` and returns nothing, similar to `Vec::push`.

src/datatype.rs
FlixCoder: Was easily possible, nice, added ^^

tests/test.rs
```diff
@@ -122,14 +126,71 @@ fn issue_9() {
...
+        format: Cow::Borrowed(&format5),
+    });
+
     range_eq!(
         range,
         [
             [String("test1".to_string())],
-            [String("test2 other".to_string())],
-            [String("test3 aaa".to_string())],
+            [RichText(cell2.clone())],
```
FlixCoder, line 192: Might be possible to use `"test2 other"` by now ^^

src/style.rs
tafia: I'd naturally pick `&str` for public api to forward proof it (avoid leaking technical details and allows an eventual other type to be used wo breaking change). I feel like most people won't need the `String` (there is already an `into_text` if necessary)

src/xlsb/cells_reader.rs
```diff
@@ -11,7 +11,7 @@ use super::{cell_format, parse_formula, wide_str, RecordIter};
 pub struct XlsbCellsReader<'a> {
     iter: RecordIter<'a>,
     formats: &'a [CellFormat],
-    strings: &'a [String],
+    strings: &'a [RichText],
```
tafia, line 14: I think it is wrong to force all strings to be `RichText`. There are already some situation where memory is an issue.

src/datatype.rs
```diff
@@ -423,8 +441,9 @@ impl DataType for DataRef<'_> {
 
     fn get_string(&self) -> Option<&str> {
         match self {
-            DataRef::String(v) => Some(&**v),
-            DataRef::SharedString(v) => Some(v),
+            DataRef::String(v) => Some(v),
+            DataRef::RichText(v) => Some(v.text()),
+            DataRef::SharedString(v) => Some(v.text()),
```
tafia, line 446: I'd expect a `SharedString` and ` SharedRichText` ??
What is the issue exactly to force `RichText` all over the place?

tafia (commented):
Thanks a lot.
I think I still don't see the reason not to keep `DataRef::SharedString` and just add a new variant `DataRef::SharedRichText` if necessary.

src/xlsb/cells_reader.rs
FlixCoder: Do you have an idea how to handle this? We need to store either normal strings or rich text, both can be the case at any shared string. So naturally you would make an enum:
```rust
enum SharedString {
    String(String),
    RichText(RichText),
}
```
But then the enum type is at least as big as RichText, if not more. So using `&[SharedString]` does not solve the memory issue.

So how are we able to store rich text while keeping the memory layout of string in the usual case? We could make the rich text varian `RichText(Box<RichText>)`, but that has different overhead, that I am not sure if it is better than additional memory use?

FlixCoder:
I can't really progress until https://github.com/tafia/calamine/pull/441#discussion_r1633001639 . Could you elaborate a bit how you imagine things to be?

src/xlsb/cells_reader.rs
wxyangf: Maybe stick it under a feature flag? like DateTime support. I think at compile time that would allow it to be the smaller String, since the enum would only have one member unless the user enables the ff, which would then warn about the memory tradeoff?

PR #445, opened 2024-06-26
docs: fixes #436

### LINKED ISSUE #436: Example Code from Readme Doesn't Work (no method named `ok_or`)
The following code in the first example does not pass `cargo check` in calamine version 0.24.0:

```
let mut workbook: Xlsx<_> = open_workbook(path)?;
    let range = workbook
        .worksheet_range("Sheet1")
        .ok_or(Error::Msg("Cannot find 'Sheet1'"))??;
```

Results in error:

```
No method named `ok_or` found for enum `Result` in the current scope
```

Using `rustc 1.77.2 `

mmorley0395:
Xlsx WorksheetNotFound error is simply propagated rather than using ok or. now passes cargo check


tafia:
Thanks

PR #449, opened 2024-07-08
Evaluate cell name directly instead of trying to be too smart

### LINKED ISSUE #446: Error while attempting to read formulas from simple XLSX file
Hello,

I am using Rust v 1.79 on M2 MacBook Pro and testing Calamine v 0.25.0.

While attempting to read the formulas from the simple attached XLSX file, the following error occurs:

error while getting formula: Xlsx(RangeWithoutColumnComponent)

This was working with previous version 0.23.1.
 
The code, directly taken from the GitHub page examples, is:

```
use calamine::{Reader, open_workbook_auto, Data};

fn main() {
    // opens a new workbook
    let path = "book_nok.xlsx"; // we do not know the file type
    let mut workbook = open_workbook_auto(path).expect("Cannot open file");

    // Read whole worksheet data and provide some statistics
    if let Ok(range) = workbook.worksheet_range("Sheet1") {
        let total_cells = range.get_size().0 * range.get_size().1;
        let non_empty_cells: usize = range.used_cells().count();
        println!("Found {} cells in 'Sheet1', including {} non empty cells",
                total_cells, non_empty_cells);
        // alternatively, we can manually filter rows
        assert_eq!(non_empty_cells, range.rows()
            .flat_map(|r| r.iter().filter(|&c| c != &Data::Empty)).count());
    }

    // Now get all formula!
    let sheets = workbook.sheet_names().to_owned();
    for s in sheets {
        println!("found {} formula in '{}'",
                workbook
                    .worksheet_formula(&s)
                    .expect("error while getting formula")
                    .rows().flat_map(|r| r.iter().filter(|f| !f.is_empty()))
                    .count(),
                s);
    }
}
```

The XLSX file is attached.

I hope this helps.

[Book_nok.xlsx](https://github.com/user-attachments/files/16049308/Book_nok.xlsx)

tafia:
`valid_cell_name` function was buggy. It turns out it may not be faster than a simple evaluation of the *cell_name* which should error-out quickly most of the time anyway.

Also simplify a little all the tests which are opening workbooks from the same tests directory.

Close #446


ajouanjean:
Hello,
Is this included in the latest release?
Looks like the error #446 is still here though.
Thanks

Issue #147, opened 2019-03-31
Range.rows() skips nonempty rows, no way to get absolute index of a row


TimoFreiberg:
I have to read excel sheets with headers where the header row is not the first row.
I currently try to solve this by first using `range.rows().enumerate().find(...)` to get a row that contains the header names I expect and then using the index of that row to get a subrange of my original range.

This doesn't work when there are empty rows before the header row.
I guess this I can currently only implement this by iterating over the cells manually, right?

tafia:
Sorry I probably do not understand your error.

Are you using [Range::range](https://docs.rs/calamine/0.15.3/calamine/struct.Range.html#method.range) method?
Are you then deserializing with this subrange?

TimoFreiberg:
Yeah. This is about usability/ergonomics.

Since my header row wasn't the first row, I had to first find the correct row and then skip `n` rows, where `n` was the index of the header row when iterating over `Range.rows().enumerate()`.

Having to find the header row manually was a bit annoying, but I'm not sure how the library would offer this functionality in a natural way.

Then I had to skip `n` rows, but I didn't notice that only nonempty rows would be returned by `Range.rows()`, so I called `my_range.range((n, 0), my_range.end().unwrap())`, although instead of `n` I should have used `my_range.start().unwrap().0 + n` (i.e. `skip n rows`).

I might not have made the second mistake if `Range` had a `skipRows(self, n: u32) -> Range` method because I wouldn't have noticed the error in my mental model of the `Rows` iterator.
I also wouldn't have had to write this helper ;)
```rust
fn skip_rows(range: Range<DataType>, n: u32) -> Result<Range<DataType>, Error> {
    if range.is_empty() {
        return Err(Error::new("Range was empty"));
    }
    let start = range.start().unwrap();
    let end = range.end().unwrap();
    Ok(range.range((start.0 + n, start.1), end))
}
```

tafia:
> but I didn't notice that only nonempty rows would be returned by `Range.rows()`.

It is not the case. Do you have an example?

I think what you're confused about is probably relative (`usize`)/absolute (`u32`) position. `Range::range` uses absolute index so indeed you should build that from `my_range.start`.

What we could do is probably improve doc first. Then we could have a new `Index<(Range<usize>, Range<usize>)>` impl which would then use relative positions.

TimoFreiberg:
Sorry for the long silence

> It is not the case. Do you have an example?
[...]

PR #453, opened 2024-07-20
feat: add option to set header row

PrettyWood:
We have an issue https://github.com/ToucanToco/fastexcel/issues/209 that has been getting some popularity
People want to be able to set the "header row" knowing where it starts and the fact that calamine skips by default empty rows at the beginning is not always wanted.

I suggest we add an option to be able to change the default behavior directly in calamine. It will be a bit less performant for the ones who enable this option since the whole vec of cells might be shifted but has no impact on people who keep the default behavior.

Note: I decided to tackle only xlsx files for now because
1) I don't know if this will be accepted
2) I don't know if the behavior is the same for other files
3) I don't know if the "XlsxOptions" is something we want to generalize as an associated type `type Option` in the `Reader `trait

related to https://github.com/tafia/calamine/issues/147#issuecomment-492716541


dimastbk:
> I don't know if the behavior is the same for other files

Yes, the same. See first four tests https://github.com/dimastbk/python-calamine/blob/master/tests/test_base.py#L11

src/xlsx/mod.rs
```diff
@@ -197,6 +197,29 @@ pub struct Xlsx<RS> {
...
     merged_regions: Option<Vec<(String, String, Dimensions)>>,
+    /// Reader options
+    options: XlsxOptions,
+}
+
+/// Xlsx reader options
+#[derive(Debug, Default)]
+pub struct XlsxOptions {
+    /// By default, calamine skips empty rows until a nonempty row is found
+    pub keep_first_empty_rows: bool,
```
tafia, line 208: I think a more generic approach would be to have the option to set header row?
By default `None` means that it'll skip all empty rows (like today) and then we could have `with_header_row(self, row: u32)`?
I understand that >90% of use cases are the first row but it still feels better and not necessarily more complicated to give the option to the users?

tests/test.rs
```diff
@@ -1786,3 +1787,39 @@ fn test_ref_xlsb() {
         ]
     );
 }
+
+#[rstest]
```
tafia, line 1791: Didn't know about rstest. It seems neat.

tafia (approved):

tafia:
>    I don't know if this will be accepted

I don't see anything wrong here. I'd like the comments to reflect where/how it applies if possible, in particular if there is no option for other files.

>    I don't know if the behavior is the same for other files

While I'd love all files to be supported equally, I understand people focusing on their own issues first. If you don't mind trying to check other files i'd be great, else it is ok. Thanks anyway for your time!

>    I don't know if the "XlsxOptions" is something we want to generalize as an associated type type Option in the Reader trait

I don't really know too. I suppose it could be something useful in more contexts but I haven't felt a strong need either.

src/xlsx/mod.rs
PrettyWood: Perfect will do that

Tim-Kracht: For what it's worth, I prefer the option to load the empty rows before the header row. I am processing unknown/volatile files and I know what row the header should be, what the column names I am expecting, and I do my own processing to resolve duplicate column names. My desire to be able to load the file as-is, figure out where the headers are, possibly report that they were on a different row than expected, identify headers on the fly as needed, etc. Sometimes there are rows before the header that are empty, sometimes there are rows that are not empty.

PrettyWood: Done

PrettyWood:
I'll try to tackle other file formats by the end of the week

PrettyWood:
@tafia I added the support for ODS. I reckon having the option at workbook level is not appropriate. We could want to open a sheet with one header row and another sheet with another header row.
For me we should have
```rs
/// Options for reading data
pub struct ReadDataOptions {
    /// header row
    pub header_row: Option<u32>,
    ...
}

pub trait Reader<RS>: Sized
where
    RS: Read + Seek,
{
...
    fn worksheet_range(&mut self, name: &str, options: ReadDataOptions) -> Result<Range<Data>, Self::Error>;
    ...
 
    fn worksheet_range_at(&mut self, n: usize, options: ReadDataOptions) -> Option<Result<Range<Data>, Self::Error>> {
        let name = self.sheet_names().get(n)?.to_string();
        Some(self.worksheet_range(&name, options))
    }
...
}

pub trait ReaderRef<RS>: Reader<RS>
where
    RS: Read + Seek,
{
    fn worksheet_range_ref<'a>(&'a mut self, name: &str, options: ReadDataOptions)
        -> Result<Range<DataRef<'a>>, Self::Error>;

    fn worksheet_range_at_ref(&mut self, n: usize, options: ReadDataOptions) -> Option<Result<Range<DataRef>, Self::Error>> {
        let name = self.sheet_names().get(n)?.to_string();
        Some(self.worksheet_range_ref(&name, options))
    }
}
```

The issue is that it would break the API
I could add new method `worksheet_range_with_option` for example but it sounds quite heavy.
WDYT?

ACronje:
@PrettyWood would your solution work for this use case?

image

PrettyWood:
@ACronje yup!

PrettyWood:
@ACronje yup!

ACronje:
Even though my Rust knowledge is pretty limited I would be willing to help get this over the line @PrettyWood @tafia .

We _really_ need this!

PrettyWood:
The rust part is not really a problem and I can finish the job easily, it's more a problem of API ;)
I need @tafia opinion on https://github.com/tafia/calamine/pull/453#issuecomment-2358655100

ACronje:
> @tafia I added the support for ODS. I reckon having the option at workbook level is not appropriate. We could want to open a sheet with one header row and another sheet with another header row. For me we should have
> 
> ```rust
> /// Options for reading data
> pub struct ReadDataOptions {
>     /// header row
>     pub header_row: Option<u32>,
>     ...
> }
> 
> pub trait Reader<RS>: Sized
> where
>     RS: Read + Seek,
> {
> ...
>     fn worksheet_range(&mut self, name: &str, options: ReadDataOptions) -> Result<Range<Data>, Self::Error>;
>     ...
>  
>     fn worksheet_range_at(&mut self, n: usize, options: ReadDataOptions) -> Option<Result<Range<Data>, Self::Error>> {
>         let name = self.sheet_names().get(n)?.to_string();
>         Some(self.worksheet_range(&name, options))
>     }
> ...
> }
> 
> pub trait ReaderRef<RS>: Reader<RS>
> where
>     RS: Read + Seek,
> {
>     fn worksheet_range_ref<'a>(&'a mut self, name: &str, options: ReadDataOptions)
>         -> Result<Range<DataRef<'a>>, Self::Error>;
> 
>     fn worksheet_range_at_ref(&mut self, n: usize, options: ReadDataOptions) -> Option<Result<Range<DataRef>, Self::Error>> {
>         let name = self.sheet_names().get(n)?.to_string();
>         Some(self.worksheet_range_ref(&name, options))
>     }
> }
> ```
> 
> The issue is that it would break the API I could add new method `worksheet_range_with_option` for example but it sounds quite heavy. WDYT?

@PrettyWood since adding options like you've suggested above would be a breaking change, and since adding a new method like `worksheet_range_with_option` would inevitably require you to add `worksheet_formula_with_options` etc. how about rather passing the sheet-specific parsing options to `with_options`? 

I think this could be useful for the other file types that do not support lazy iteration (they are loading the sheets into memory at instantiation time I believe? see: [ods](https://github.com/tafia/calamine/blob/master/src/ods.rs#L152C13-L152C26)) and therefore might decide to use this sheet-specific config to, for example, be more efficient with loading sheet data.

PrettyWood:
I first went with the `with_options` approach (it's still the case in this PR) but I reckon it's wrong to have sheet options (like `header_row`) that are meant for a single sheet to be set at workbook level.

ACronje:
I understand what you're saying and I saw your implementation. I explained why I think it's acceptable for the sheet options to be at the workbook level.

PrettyWood:
Thanks for sharing your opinion. I'll finish implementing all the extensions and let the `with_options` for now.
If it's good enough it will be mergeable directly. If not it won't cost too much to update

severinh:
Huge thanks to @PrettyWood for this improvement! Our team depends on this PR to fully adopt calamine.

@tafia Even though you've already approved the PR, I believe @PrettyWood is still waiting for you to take another look before merging. Any chance you could take a look in the coming days? Or can he go ahead and merge? Thanks a lot!

src/auto.rs
```diff
@@ -74,17 +75,57 @@ where
...
+            Sheets::Xlsb(ref mut e) => match options {
+                AutoReaderOptions::Xlsb(opts) => e.set_options(opts),
+                _ => unreachable!(),
+            },
+            Sheets::Ods(ref mut e) => match options {
+                AutoReaderOptions::Ods(opts) => e.set_options(opts),
+                _ => unreachable!(),
+            },
+        }
+    }
```
tafia, line 127:
```suggestion
    fn set_options(&mut self, options: Self::Options) {
        match (*self, options) {
            (Sheets::Xls(ref mut e), AutoReaderOptions::Xls(opts)) => e.set_options(opts),
            (Sheets::Xlsx(ref mut e), AutoReaderOptions::Xlsx(opts)) => e.set_options(opts),
            (Sheets::Xlsb(ref mut e), AutoReaderOptions::Xlsb(opts)) => e.set_options(opts),
            (Sheets::Ods(ref mut e), AutoReaderOptions::Ods(opts)) => e.set_options(opts),
            _ => unreachable!(),
        }
    }
```

tafia (approved):

tafia:
Thanks a lot for the extra effort! LGTM, just one tiny refactor I believe makes the intent clearer.

src/auto.rs
PrettyWood: I first tried that but I had the `move occurs because `*self` has type `Sheets<RS>`, which does not implement the `Copy` trait`
But in fact we can just switch to maching on reference itself! And it should be faster :)

PrettyWood: [dd33174](https://github.com/tafia/calamine/pull/453/commits/dd33174899447bdb02d34b9eee622eeafd0eb9bd)

tafia:
@PrettyWood  let me know once you think you're done. I'm happy to merge it as it is already.
Thanks again!

PrettyWood:
Ok @tafia sorry for those last changes but I prefer switching to an enum. It's more explicit and I plan on adding `skip_rows` and `n_rows` to options after. So I guess it will make sense to also have `None` for the header row to say "I don't want any header at all". Right now there is no difference between "the first row of the range" and "the header row" but it will make sense after if some kind of pagination is added

tafia:
Awesome thanks!

PR #456, opened 2024-08-10
fix(ods): support consecutive repeated empty cells

### LINKED ISSUE #455: Strange failure to determine `string` dtype column in ODS sheet/file
Original issue: https://github.com/ToucanToco/fastexcel/issues/275

PrettyWood:
closes #455 

### Explanation

We had two files: `single-empty.ods` that worked fine and `multi-empty.ods` that didn't
Why? Because for `single-empty.ods` the XML behind it only has one empty with `table:table-row table:number-rows-repeated="7"` whereas `multi-empty.ods` has one empty, one with `table:table-row table:number-rows-repeated="4"` and then one with `table:table-row table:number-rows-repeated="2"`

The code didn't support consecutive empty lines. I tried to make as little change as possible but there might well be a better way


alexander-beedie:
Nice one.

archqt:
A new version 1.5.0-1 is available for me on Archlinux, and still the same it doesn't work.

tafia:
Thanks!

PR #458, opened 2024-08-16
docs: enable dates and picture features in docs.rs build

paolobarbolini:
While looking at the docs through docs.rs I didn't realize `ExcelDateTime` is already able to be converted into `chrono::NaiveDateTime` because the `dates` feature is not enabled in the build.
This PR leverages the [docs.rs metadata feature](https://docs.rs/about/metadata) to enable the `dates` and `picture` features when building docs for docs.rs.


tafia:
Thanks!

PR #462, opened 2024-08-29
Missing text

sftse:


sftse:
Based on #457

sftse:
For some reason this example exhibits a record type that I've not been able to locate in the spec, do you have any idea what it could be?
This was found by differentially testing .xls converted to .xlsx with the help of Libreoffice and extracting the text with calamine from both .xls and .xlsx.
The dataset are .xls from the Enron email dataset, dated around 2000.

sftse:
Folded into #463

PR #463, opened 2024-08-29
Oom alloc

sftse:
Was it intentional that `fn from_sparse` assumes cells are in `(row, col)` order? This test example demonstrates this is not always the case and OOM allocates due to an overflow.

Based on #462


sftse:
Added another example that OOMs, hasn't been fixed yet, I'm working around it for my usecase by using unstable `set_alloc_error_hook`

tafia:
Thanks!
Tiny comment just to help future debugging if needed.

sftse:
Do you mean a comment explaining the last OOM file that was added?

sftse:
I've added a FIXME to the function call that OOMs, if this was not what was meant please advise further.
This is ready to merge imho.

sftse:
Curious whether you know anything about the unknown record types as mentioned in #462.
The commit history was starting to look complex, so I folded it into this PR.

tests/test.rs
```diff
@@ -2158,6 +2158,12 @@ fn test_high_byte_strings_and_unicode_strings_without_reserved_tags() {
             }
         }
     }
+    // FIXME: Libreoffice recognizes a REPT("O", I44) formula
```
jmcnamara, line 2161: Do you have a sample file with the unrecognised tag?

src/xls.rs
```diff
@@ -490,7 +490,15 @@ impl<RS: Read + Seek> Xls<RS> {
                                 });
                         formulas.push(Cell::new(fmla_pos, fmla));
                     }
-                    _ => (),
+                    // tests/high_byte_string.xls contains a record type that
+                    // cannot be found in the "By Number" 2.3.2 table
+                    0x00D6 => {
```
jmcnamara, line 495: Tag 0x00D6 is `RSTRING`, i.e., rich multi-format strings.

tests/test.rs
sftse: The assert below attempts to retrieve the value, it does not recognize the formula, is this not what you meant?

jmcnamara: I was asking if you have a test file with this tag?

jmcnamara: 192 is 0x00C0 = 'TOOLBAREND'. I'm just not clear if that is the context it is appearing in here.

It would normally be accompanied by 0x00BF => 'TOOLBARHDR'.

sftse: This set of commits introduced this test and the test file.

src/xls.rs
jmcnamara: It doesn't seem to be mentioned in the Microsoft online docs (I have an old paper copy of the "Excel Developer's  Kit" and it is in that. Next best to link to would be the OpenOffice docs for the "Microsoft Excel File Format": https://www.openoffice.org/sc/excelfileformat.pdf

src/lib.rs
```diff
@@ -722,7 +722,9 @@ impl<T: CellType> Range<T> {
     /// assert_eq!(range.used_cells().count(), 3);
     /// ```
     ///
-    pub fn from_sparse(cells: Vec<Cell<T>>) -> Range<T> {
+    pub fn from_sparse(mut cells: Vec<Cell<T>>) -> Range<T> {
+        // cells do not always appear in (row, col) order
+        cells.sort_by_key(|cell| (cell.pos.0, cell.pos.1));
```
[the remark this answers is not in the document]
sftse: No longer sorts in the improved PR #524.

PR #464, opened 2024-09-03
feat: named table data as `DataRef` type

wdoppenberg:
During some work I was performing on a MR for `fastexcel` (https://github.com/ToucanToco/fastexcel/pull/282) I noticed no method for retrieving `Table`s with `DataRef` as inner type was not possible. This MR should hopefully resolve this issue. 

I also implemented `From<Table<T>>` for `Range<T>` to prevent large clone operations, which I ran into in one method in particular inside `fastexcel`. 

Finally I resolved a small `clippy` warning mentioning that `std::u32::MAX` is deprecated. 

Eager to receive your feedback :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -955,6 +955,11 @@ impl<T> Table<T> {
     pub fn data(&self) -> &Range<T> {
         &self.data
     }
+
+    /// Get an owned range representing the data from the table (excludes column headers)
+    pub fn data_owned(self) -> Range<T> {
+        self.data
+    }
 }
 
 /// A helper function to deserialize cell values as `i64`,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -932,13 +932,13 @@ impl<'a, T: 'a + CellType> DoubleEndedIterator for Rows<'a, T> {
 impl<'a, T: 'a + CellType> ExactSizeIterator for Rows<'a, T> {}
 
 /// Struct with the key elements of a table
-pub struct Table<T: CellType> {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -932,13 +932,13 @@ impl<'a, T: 'a + CellType> DoubleEndedIterator for Rows<'a, T> {
 impl<'a, T: 'a + CellType> ExactSizeIterator for Rows<'a, T> {}
 
 /// Struct with the key elements of a table
-pub struct Table<T> {
+pub struct Table<T: CellType> {
```
tafia, line 935: I think it is better to refrain from adding constraints on the structs, the impl are enough, more generic and more future proof.

src/lib.rs
```diff
@@ -932,13 +932,13 @@ impl<'a, T: 'a + CellType> DoubleEndedIterator for Rows<'a, T> {
...
     pub(crate) name: String,
     pub(crate) sheet_name: String,
     pub(crate) columns: Vec<String>,
     pub(crate) data: Range<T>,
 }
-impl<T> Table<T> {
+impl<T: CellType> Table<T> {
```
tafia, line 941: Same here, it is unnecessary I think.

tafia (approved):
Thanks!
Small comments to keep it as generic as possible.

wdoppenberg:
@tafia Is there anything else you need me to change in this PR?

tafia:
Sorry for the late answer lgtm!

PR #467, opened 2024-09-18
Ignore Annotations

schrieveslaach:
With this commit annotations will be ignore in the parsing process to make sure the actual values will be returned only.


tafia:
Thanks!

schrieveslaach:
Thanks for accepting this PR. :blush: 

Just a question: do you plan a release of the crate soon? I would like to upgrade it in nushell so that the functionality will be available in my shell. (CC: @fdncred)

PR #470, opened 2024-10-08
fix: header row in the middle can break

PrettyWood:
https://github.com/tafia/calamine/pull/453 had a bug if the header row was not 0 and was on an empty line.
We would insert an empty cell in the middle of `cells` vec, which would make `Range::from_sparse(cells)` break because it requires `cells` to be sorted


tafia:
Thanks!

tafia:
bumped v0.26.1

PR #472, opened 2024-10-20
fix: always parse a string cell as string

### LINKED ISSUE #266: some string formula cells are lost
First off thanks for the shoutout in the credits!

https://github.com/SheetJS/test_files/raw/master/large_strings.xlsx the first worksheet was constructed using string concatenation.  Using the `excel_to_csv` example in the repo, the generated output is dropping all of the formula cells like

```xml
      <c r="A2" t="str">
        <f t="shared" ref="A2:Q2" si="0">CONCATENATE(A1,"a")</f>
        <v>aa</v>
      </c>

      <c r="A6" t="str">
        <f t="shared" si="1"/>
        <v>aaaaaa</v>
      </c>
```

There's a strange comment:

```
                // NB: the result of a formula may not be a numeric value (=A3&" "&A4).
                // We do try an initial parse as Float for utility, but fall back to a string
                // representation if that fails
```

Is this needed?  A formula with a numeric result will be marked with cell type `n`.  Even in the posited case where A3 is 1 and A4 is 2 and `A5=A3&" " &A4

The linked website in the comment is misleading.  It is better to reference the spec:

ST_CellType

The `str` cell type exclusively refers to a formula whose result is a string.  The `v` child holds the string value of the result.

PrettyWood:
Fix issue coming from https://github.com/pola-rs/polars/issues/19169

Comment from https://github.com/tafia/calamine/issues/266#issue-1322822362 seems to indicate there should not be any reason to try to parse a "str" cell as float since "a formula with a numeric result will be marked with cell type n"

Closes #266


tafia:
Thanks!

PR #473, opened 2024-10-20
build(deps): bump quick-xml to 0.36

### LINKED ISSUE #454: bump quick-xml dependency
Since quick-xml 0.31 was released late 2023, there have been six releases, all within the last two months.

https://crates.io/crates/quick-xml/versions

calamine only compiles to 0.31.0 as there are breaking API changes 0.32 and above.

Are there any benefits that calamine can benefit from in later release of quick-xml?

cc @mingun

PrettyWood:
fix #454


tafia:
Thanks!

PR #479, opened 2024-11-12
Malformed format

sftse:


tafia:
Hi do you mind writing a description of what was wrong exactly? I can read the code but it makes understanding high level issues better when they are written out.
Thanks

sftse:
The code reads at least 5 bytes from the slice when parsing, but the error check only covers four, leading to a panic on `r.data[4]`.
Since this is the only thing preventing the provided example case from parsing, continue when such errors are encountered.

This panic could be explained differently, if a defective record is passed to `fn parse_format`, but record parsing being correct seems so fundamental to understanding xls that it seemed unlikely.

Issue #378, opened 2023-11-29
Use atoi_simd for SIMD accelerated &[u8] to integer conversion


jqnatividad:
The polars project uses it heavily and it should give a noticeable performance boost.

https://crates.io/crates/atoi_simd

jqnatividad:
resolved by #482

PR #482, opened 2025-01-01
perf: accelerated datatype conversion

jqnatividad:
implements #378

Apart from `atoi_simd` for integer conversion, also used `fast-float2` for float conversion.

Both crates are heavily used by the Polars project.


tafia:
Thanks!

PR #490, opened 2025-03-12
read_string: fix comparison with 'closing' tag name

falconandy:
One of my xlsx file contains namespace-prefixed `si` tag name (`x:si`)  in `xl/sharedStrings.xml`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<x:sst count="1187" uniqueCount="1187" xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
    <x:si>
        <x:t>String 1</x:t>
    </x:si>
...
>
```
Without the fix I get a error `Xlsx error: Unexpected end of xml, expecting '</>'` when I try to open the file - the code tries to compare namespace-prefixed tag name with local tag name.

I am a Rust novice - maybe the fix/test code is not idiomatic.


tafia (approved):

tafia:
Thanks

PR #497, opened 2025-04-08
Pin zip crate to 2.5.*

prophittcorey:
This fixes the current breaking build due to zip's 2.6 breaking changes.

Here's a screenshot of the build issues due to zip >= 2.6.

image


tafia:
Thanks!

PR #521, opened 2025-07-02
Byte buffers holding XLUnicodeString prematurely truncated

sftse:
This PR is another batch of commits from #463.

The first commit b5b948a6caa6c23708e2bc26d8fc1e322b7c4126 introduces a test that fails.
1c58893cff3066d10b00afd8caeeaede098420a0 fixes one of the mistakes, ~~so that a different error message appears.
599eae8ce442a922b3ccf8982471f91d143fac3f fixes the next failure due to a spec nonconformity of the test file, which seems like a logical extension of the standard.~~ Already fixed in upstreamed PR.

Edit: updated commit hashes after push


sftse:
Before merge I want to add the particular, spec-nonconformant `XLUnicodeString` as a unit test.

sftse:
While revisiting this to add a unit test, I noticed that the formula that is triggering this high byte string panic is also the one from row 43 col 9 that was the point of discussion [here](https://github.com/tafia/calamine/pull/463/commits/869edb7196567b66135486ec54b5183055de641d#r2182455192).

This makes me think there is more to this test than initially thought. The other thing I'm not able to piece together is that the `XlsEncoding` is `windows-1252`, so the high-byte being set for two byte characters doesn't make a lot of sense? The root cause for this error might be something else, and the patch does fixup an error, just a different one, while accidentally fixing the crash.

If there are any doubts about how the `MS-XLS 2.5.296` says one should read an `XLUnicodeStringNoCch`, then we should hold off merging, if not I think we can merge? A testcase would be nice though, since I'm no longer certain the .xls file added even contains any.

sftse:
Actually, let's hold off on merge until we can investigate this further.

jmcnamara:
> While revisiting this to add a unit test, I noticed that the formula that is triggering this high byte string panic is also the one from row 43 col 9 that was the point of discussion [here](https://github.com/tafia/calamine/pull/463/commits/869edb7196567b66135486ec54b5183055de641d#r2182455192).


I created a test xls file called `j44.xls` that contains the formula `=REPT("O",I44)` in cell J44 like in the failing test file `high_byte_string.xls`.

[j44.xls](https://github.com/user-attachments/files/21090028/j44.xls)


This has the following binary data for the formula:

```
    17 01 00 4F 44 2B 00 08 C0 41 1E 00
```

I intrepreted this as:

```
17    ptgStr
01    len
00    cch = single byte
4F    string "O"

44    ptgRefV
002B  Row 43
08    Col  8 (combined to make I44)

C0    Separator??

41    ptgFuncV
001E  REPT
```

(As an aside, I said in another comment that `C0` was the record "TOOLBAREND" but I realise now that you were probably asking in a ptg context)

The formula in `high_byte_string.xls` looks like this?

```
   17 01 4F 44 2B C0 08 41 1E 00 
```

Comparing the 2 binary data sets:

```
high_byte: 17 | 01    4F | 44 2B C0 08 |    | 41 1E 00

j44:       17 | 01 00 4F | 44 2B 00 08 | C0 | 41 1E 00
```

There are 2 odd things here.

1. There is no `fHighByte` field in the the `ShortXLUnicodeString`. This may be because, as you point out, the CODEPAGE is 1252 which might mean that all strings are simple single byte encodings so the encoding/fHighByte field can be omitted.
2. The row in the cell reference appears to be a single byte only. And the `C0` token is in an odd place.



All in all, it looks like a junk file except for the fact that Excel (including a couple of older versions that I have) reads it. This is one of the reasons I stopped working with the xls format once xlsx came along. :-)


We could maybe just fix the `cch = 1` issue and not fix whatever this is.

jmcnamara:
@sftse I undated the previous comment to clarify what `C0` is.

jmcnamara:
@sftse Is this patchset ready for merge?

jmcnamara:
Let me know if you still plan to do this work or if it can be merged as is.

jmcnamara:
@sftse I would like to close this until it is actively targeting upstream. I am trying to reduce the PR backlog to items that are in progress.

sftse:
Agree, no point having this open forever.

jmcnamara:
Thanks for understanding.

PR #525, opened 2025-07-07
More OOM fixes

sftse:
This is another batch of commits from #463.

This adds another test case that OOM allocates.
I rechecked the work and noticed that the `fn parse_dimensions` is in some sense superfluous. This fixes the function as-is, but I don't see what the function is even doing, or if there are any features that depend on it. It only is used to reserve space on the `Vec` holding the cells, and from basic testing always overallocates, sometimes by quite a bit.

If no features depend on parsing the dimensions it might be sensible to remove, but we can also merge as-is.


jmcnamara:
LGTM. I ran Copilot review since it sometimes picks up something useful. Not sure the "named constant" comment is worth changing though. I'm happy to merge as-is unless you want to change it.



> I rechecked the work and noticed that the `fn parse_dimensions` is in some sense superfluous. This fixes the function as-is, but I don't see what the function is even doing, or if there are any features that depend on it. It only is used to reserve space on the `Vec` holding the cells, and from basic testing always overallocates, sometimes by quite a bit.

That would require a call by @tafia. 

I'll review it in the context of the Xlsx file when I get to that part of the docs. Let's leave it in place for now.

sftse:
Named constants make sense to me if there really is a meaningful name to attach to it or if it is used repeatedly. In this case it really is a magic constant extracted from the spec, with no obvious indication why it may not be larger than 0xFF. Referencing the spec is about as meaningful as it can get.

Good to merge.

jmcnamara:
Merged. Thanks.

PR #528, opened 2025-07-14
track Cargo.lock

### LINKED ISSUE #527: Dependency zip.rs v4.3.0 requires a MSRV of 1.85.0
Zip.rs v4.3.0 requires a MSRV of v1.85.0 whereas Calamine's stated MSRV is v1.73.0. This is causing a failure in CI and presumably for any users with rustc < v1.85.0:

```
Run cargo build
    Updating crates.io index
 Downloading crates ...
  ...
error: package `zip v4.3.0` cannot be built because it requires rustc 1.82.0 or newer, 
while the currently active rustc version is 1.75.0
Either upgrade to rustc 1.82.0 or newer, or use
cargo update zip@4.3.0 --precise ver
where `ver` is the latest version of `zip` supporting rustc 1.75.0
Error: Process completed with exit code 101.
```

https://github.com/tafia/calamine/actions/runs/16188594372/job/45698992946

sftse:
~~Fixes #527~~

Although that issue has other resolutions, tracking the lock file is the fastest resolution, with added benefits.
Knowing the exact versions that were tested by CI is valuable, as crates are known to release versions that violate semver.
Not having the lock file makes `git bisect` very frustrating, as the dependency resolution at a given commit may resolve to different versions than the ones that passed CI that are broken and don't compile. It also helps when tracking regressions to know as precisely as possible what the whole context of the build was.

OTOH, one of the drawbacks is that git handles merge conflicts on the lock file quite poorly, sometimes resulting in quite a mess.


sftse:
This is really orthogonal to #527 since it does not directly relate to the MSRV but instead ensures that successful builds are reproducible. If additional constraints related to MSRV are codified they will influence the lock file, not the other way around. This merely prevents the computed lock file from being lost.

It's unfortunate that merging this will close that issue, I can't seem to unlink it, but it can be reopened afterwards. Alternatively, I can make an identical PR that does not reference #527 at all.

sftse:
My mistake for referencing #527 at all. This really is a fix for the broken CI.

Issue #530, opened 2025-07-15
bug:  Xlsx error: File not found 'xl/_rels/workbook.xml.rels'


vbingbingv:
### Question

Does this mean there is something wrong with the file format? I am using a Windows computer and I can open it with Office.

jmcnamara:
Please upload the file that causes the warning.

jmcnamara:
The issue is that the xlsx/zip file uses Windows style directory separators so `xl\_rels\workbook.xml.rels` exists but `xl/_rels/workbook.xml.rels` doesn't exist, and hence the error.

This is somewhat related to this issue in `zip.rs`: https://github.com/zip-rs/zip2/issues/272

I'll look into a fix with normalized file paths. 

Out of curiosity, what software created this xlsx file?

vbingbingv:
> The issue is that the xlsx/zip file uses Windows style directory separators so exists but doesn't exist, and hence the error.`xl\_rels\workbook.xml.rels``xl/_rels/workbook.xml.rels`
> 
> Out of curiosity, what software created this xlsx file?
> 
> This is somewhat related to this issue in : [zip-rs/zip2#272](https://github.com/zip-rs/zip2/issues/272)`zip.rs`
> 
> I'll look into a fix with normalized file paths. Simpler file to reproduce:
> 
> use calamine::{open_workbook, Error, Xlsx};
> 
> fn main() -> Result<(), Error> {
>     let path = "62aqrxjsin9o.xlsx";
> 
>     let _: Xlsx<_> = open_workbook(path)?;
> 
>     Ok(())
> }

thks~
All of them are xlsx files generated by office software. Now the situation is that I have a bunch of excel files locally, and I need to parse out the contents of these files. However, I can't parse some of them, but I can parse others

jmcnamara:
> All of them are xlsx files generated by office software.

Which office software? 

It isn't created by Excel because it doesn't (probably can't) use backward slashes in the part names and also the XML in the sub files aren't compact (i.e., without additional whitespace). So the `62aqrxjsin9o.xlsx` file must have been created with some other software. Do you know what that is.

jmcnamara:
> I asked some of my colleagues, and some of them also used [WPS](https://www.wps.com/) to create office documents.

Thanks. Could you ask them to create a minimal xlsx file in WPS with the one work or character in cell A1 and attach it for reference.

BTW, I've added a fix for this in #531 but it may take some days to get merged.

vbingbingv:
[...]

PR #531, opened 2025-07-15
Add tests and fixes for issues with Excel packaging part names

jmcnamara:
- test: add test for issue #419

  Added test for issue #419 where the part name is sentence case
  instead of camel case. The test file contains a sub-file
  called "xl/SharedStrings.xml" (note the uppercase S in Shared).
  This is allowed by "Office Open XML File Formats — Open Packaging
  Conventions" 6.2.2.3.

 - test: add test for issue #530

   Added test for issue #530 where the Excel packaging part names
   contain backslashes as the directory separator.
  
   This may not be stricly allowed according to "Office Open XML
   File Formats — Open Packaging Conventions" 6.2.2.2 but it turned
   up in files in the wild.

 - xlsb: add test and fix for issue #419


jmcnamara:
Note on issue #530 where the failing file has Windows style separators like "xl\_rels\workbook.xml.rels" instead of "xl/_rels/workbook.xml.rels". Strictly speaking this doesn't seem to be allowed by the Office Open XML File Formats - Open Packaging Conventions":

>### 6.2.2 Part names
>#### 6.2.2.1 General
>A part shall have a part name, which shall uniquely identify a part within an abstract package.
>#### 6.2.2.2 Syntax
>A part name shall be a Unicode string that matches the following production rules in the ABNF syntax
defined in RFC 5234

    part_name = 1*( "/" isegment-nz )
    isegment-nz = <isegment-nz, see RFC3987, Section 2.2>

>and that further satisfies the constraints listed below, where an I18N segment is a Unicode string that
matches the non-terminal isegment-nz and percent-encoding represents a character by the percent
character "%" followed by two hexadecimal digits, as specified in RFC 3986
>- No I18N segments shall contain percent-encoded forward slash (“/”), or backward slash (“\”)
characters.

However, it is probably just easier to fix it than to reject is as not conforming to the spec. Especially since a related normalization is required to allow for case-insensitive path/part names.

PR #538, opened 2025-07-24
Implement styles

ddimaria:
Add style support for Calamine.

Basic api for getting styles
```rust
fn test_worksheet_style_iter() {
    let mut xlsx: Xlsx<_> = wb("styles.xlsx");
    let styles = xlsx.worksheet_style("Sheet 1").unwrap();

    for (row, styles) in styles.rows().enumerate() {
        for (col, style) in styles.iter().enumerate() {
            println!("row: {}, col: {}, style: {:?}", row, col, style);
        }
    }
}
```

Here is the relevant structs so far:
```rust
struct Style {
    font: Option<Font>,
    fill: Option<Fill>,
    borders: Option<Borders>,
    alignment: Option<Alignment>,
    number_format: Option<NumberFormat>,
    protection: Option<Protection>,
    style_id: Option<u32>,
}

struct Color {
    alpha: u8,
    red: u8,
    green: u8,
    blue: u8,
}

struct Borders {
    left: Border,
    right: Border,
    top: Border,
    bottom: Border,
    diagonal_down: Border,
    diagonal_up: Border,
}

struct Font {
    name: Option<String>,
    size: Option<f64>,
    weight: FontWeight,
    style: FontStyle,
    underline: UnderlineStyle,
    strikethrough: bool,
    color: Option<Color>,
    family: Option<String>,
}

struct Alignment {
    horizontal: HorizontalAlignment,
    vertical: VerticalAlignment,
    text_rotation: TextRotation,
    wrap_text: bool,
    indent: Option<u8>,
    shrink_to_fit: bool,
}

struct Fill {
    pattern: FillPattern,
    foreground_color: Option<Color>,
    background_color: Option<Color>,
}

struct NumberFormat {
    format_code: String,
    format_id: Option<u32>,
}

struct Protection {
    locked: bool,
    hidden: bool,
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "calamine"
-version = "0.34.0"
+version = "0.35.0"
 authors = ["Johann Tuffe <tafia973@gmail.com>"]
 repository = "https://github.com/tafia/calamine"
 documentation = "https://docs.rs/calamine"
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,79 @@ This is the changelog/release notes for the `calamine` crate.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.35.0] - 2026-5-10
+
+### Added
+
+- Added `has_1904_epoch()` method to Xls/Xlsx/Xlsb structs to allow the user to
+  determine which date epoch is in use by Excel. [PR #630].
+
+  [PR #630]: https://github.com/tafia/calamine/pull/630
+
+- Made `RowDeserializer` public to allow a custom `DeserializeSeed`.
+  [PR #635].
+
+  [PR #635]: https://github.com/tafia/calamine/pull/635
+
+### Changed
+
+- Added deserialization examples. [PR #649].
+
+  [PR #649]: https://github.com/tafia/calamine/pull/649
+
+- Performance improvements:
+  - Optimized `<v>` cell value parsing. [PR #622]
+  - Optimized cell reference parsing. [PR #623]
+
+  [PR #622]: https://github.com/tafia/calamine/pull/622
+  [PR #623]: https://github.com/tafia/calamine/pull/623
+
+### Fixed
+
+- Fixed issue with redundant `<v>` value in inlineStr XLSX cells. [Issue #633].
+
+  [Issue #633]: https://github.com/tafia/calamine/issues/633
+
+- Fixed issue when processing XLSX files to ignore namespace prefix when reading
+  relation ids. Previously accepted only `r:` and `relationships:` prefixes.
+  [Issue #634].
+
+  [Issue #634]: https://github.com/tafia/calamine/issues/634
+
+- Fixed XLSX issue where rich text `<r>` elements were ignored after initial
+  plain `<t>` elements. [Issue #636].
+
+  [Issue #636]: https://github.com/tafia/calamine/issues/636
+
+- Fixed XLS issue where date format was ignored for formula values.
+  [Issue #640].
+
+  [Issue #640]: https://github.com/tafia/calamine/issues/640
+
+- Fixed XLSX shared formula expansion corrupting function names. The
+  `replace_cell_names()` function treated `LOG10` as a cell reference (column
+  LOG, row 10). [Issue #644].
+
+  [Issue #644]: https://github.com/tafia/calamine/issues/644
+
+- Fixed XLSX `inlineStr` issue for elements that contain CDATA. [Issue #648].
+
+  [Issue #648]: https://github.com/tafia/calamine/issues/648
+
+- Fixed panic when parsing formulas in XLS BIFF5 workbooks.The `parse_formula`
+  function decoded the non-3D PtgRef, PtgArea, PtgRefErr, and PtgAreaErr tokens
+  assuming BIFF8 sizes. [PR #643].
+
+  [PR #643]: https://github.com/tafia/calamine/pull/643
+
+- Fixed XLS panic with unsupported formula functions. Handle unsupported XLS
+  function ids as formula parse errors instead of indexing past the function
+  table. [Issue #646].
+
+  [Issue #646]: https://github.com/tafia/calamine/issues/646
+
+
 ## [0.34.0] - 2026-03-07
 
 ### Changed
@@ -26,10 +99,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [PR #615]: https://github.com/tafia/calamine/pull/615
 
 - Performance improvements:
-  - Cell reader buffer reuse and optimised maps. [PR #611]
-  - Optimised XLS sector chain reads. [PR #609]
+  - Cell reader buffer reuse and optimized maps. [PR #611]
+  - Optimized XLS sector chain reads. [PR #609]
   - Ensured `fast_float2` is used at all float-parsing call sites. [PR #608]
-  - Cached path and attr lookups, reused buffers, minimised allocations. [PR #606]
+  - Cached path and attr lookups, reused buffers, minimized allocations. [PR #606]
 
   [PR #606]: https://github.com/tafia/calamine/pull/606
   [PR #608]: https://github.com/tafia/calamine/pull/608


azarmadr:
Is the data and styles coming together?
If not, then would we have to read the excel twice to get data, once and then style?

azarmadr:
i got it, even formulae were being used similarly. Was able to satisfy my requirement with `worksheet_sytle` over a range.

Another doubt, why every item in the style is an option?
All the fonts, borders, and others are available as ids in cellxfs?

hope i am not bothering you.😊

ddimaria:
> Another doubt, why every item in the style is an option?

@azarmadr this is because the particular style may not exist for a given cell.

ddimaria:
@jmcnamara We're currently using this fork in production and it's working well.  What do I need to do to get this merged with Calamine?  Some things to discuss:
* Is the API agreeable?
* What type of documentation is needed?
* Add more testing?
* Should this be a compiler feature flag?

examples/style_example.rs
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: MIT
```
jmcnamara, line 1: +1 for example programs.

src/datatype.rs
```diff
@@ -21,6 +22,67 @@ const EXCEL_1900_1904_DIFF: f64 = 1462.;
 #[cfg(feature = "dates")]
 const MS_MULTIPLIER: f64 = 24f64 * 60f64 * 60f64 * 1e+3f64;
 
+/// A struct that combines cell value and style information
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct CellData {
+    /// The cell value
+    pub value: Data,
+    /// The cell style
+    pub style: Option<Style>,
```
jmcnamara, line 31: Strictly speaking the style is never optional in the Excel file formats. There is always an implicit 0 style that is the default and is defined in the `theme.xml` file. Note, I'm not necessarily saying that it shouldn't be an `Option<>`.

src/datatype.rs
```diff
@@ -21,6 +22,67 @@ const EXCEL_1900_1904_DIFF: f64 = 1462.;
...
+    pub fn set_style(&mut self, style: Style) {
+        self.style = Some(style);
+    }
+
+    /// Remove the cell style
+    pub fn remove_style(&mut self) {
+        self.style = None;
+    }
+
+    /// Check if the cell has style information
```
jmcnamara, line 68: Again, an Excel cell will always have style information even if it is implicit for size reasons.

src/lib.rs
```diff
@@ -451,6 +457,39 @@ impl<T: CellType> Cell<T> {
         Cell {
             pos: position,
             val: value,
+            style: None,
+        }
+    }
+
+    /// Creates a new `Cell` instance with style information.
```
jmcnamara, line 474: This level of documentation is good.

src/lib.rs
```diff
@@ -489,6 +528,86 @@ impl<T: CellType> Cell<T> {
...
+    /// use calamine::{Cell, Data, Style, Font, FontWeight};
+    ///
+    /// let mut cell = Cell::new((1, 2), Data::Int(42));
+    /// let style = Style::new().with_font(Font::new().with_weight(FontWeight::Bold));
+    /// cell.set_style(style);
+    ///
+    /// assert!(cell.get_style().is_some());
+    /// ```
+    ///
+    pub fn set_style(&mut self, style: Style) {
```
jmcnamara, line 577: Again, why is there a setter (and remove) method.

src/xlsx/style_parser.rs
```diff
@@ -0,0 +1,469 @@
...
+            }
+            b"theme" => {
+                // TODO: Handle theme colors
+                return Ok(None);
+            }
+            b"indexed" => {
+                // TODO: Handle indexed colors
+                return Ok(None);
+            }
+            _ => {}
```
jmcnamara, line 159: You may need to consider the "Automatic" and "Default" color options.

src/style.rs
```diff
@@ -0,0 +1,768 @@
...
+    /// Convert to ARGB integer
+    pub fn to_argb(&self) -> u32 {
+        ((self.alpha as u32) << 24)
+            | ((self.red as u32) << 16)
+            | ((self.green as u32) << 8)
+            | (self.blue as u32)
+    }
+
+    /// Check if the color is transparent
+    pub fn is_transparent(&self) -> bool {
```
jmcnamara, line 55: Transparency is usually stored as a separate parameter outside the color. I don't think this would actually tell the user if the color was transparent.

src/xlsx/style_parser.rs
```diff
@@ -46,15 +57,28 @@ fn parse_color(attributes: &[Attribute]) -> Result<Option<Color>, XlsxError> {
 /// Parse font weight from string
 fn parse_font_weight(s: &str) -> FontWeight {
     match s {
-        "bold" => FontWeight::Bold,
-        _ => FontWeight::Normal,
+        "bold" | "700" => FontWeight::Bold,
```
jmcnamara, line 168: I think technically in the Xls format bold can have other weights but I haven't seen it in practice.

src/style.rs
```diff
@@ -603,6 +603,195 @@ impl Protection {
...
+        self.column_widths.iter().find(|cw| cw.column == column)
+    }
+
+    /// Get row height for a specific row
+    pub fn get_row_height(&self, row: u32) -> Option<&RowHeight> {
+        self.row_heights.iter().find(|rh| rh.row == row)
+    }
+
+    /// Get effective column width (custom or default)
+    pub fn get_effective_column_width(&self, column: u32) -> f64 {
```
jmcnamara, line 775: This effective width is probably not the width that the user would expect and also may be dependent on the default font for the file.

See the following for the calculation in the other direction: https://github.com/jmcnamara/rust_xlsxwriter/blob/main/src/worksheet.rs#L19340-L19368

However, there may not be any simple way to return the correct value in all cases.

src/style.rs
```diff
@@ -603,6 +603,195 @@ impl Protection {
...
+        self.get_column_width(column)
+            .map(|cw| cw.width)
+            .or(self.default_column_width)
+            .unwrap_or(8.43) // Excel default column width
+    }
+
+    /// Get effective row height (custom or default)
+    pub fn get_effective_row_height(&self, row: u32) -> f64 {
```
jmcnamara, line 783: Same comment for row height although the height calculation is simpler.

jmcnamara:
First review done. Some general comments:

* Please squash and or rebase WIP code and reverts into a small number of logical commits or one commit.
* Fix the clippy and or format warnings the are breaking the CI.
* Add a full stop/period at the end of all comments.
* Avoid /// comments for non-public non-doc comments. Use // instead.
* Post an explanation here for the setting functions in `CellData` and `Style`.
*

ddimaria:
> @ddimaria Any update on this?

@jmcnamara sorry for the delay, I got pulled into other projects.  My goal is to polish this PR and address feedback next week.  We've been using the fork in production, so the base of the code is working for us.  As a stretch goal I'd like to implement a better storage structure to account for repeating elements (RLE).

jmcnamara:
> My goal is to polish this PR and address feedback next week. We've been using the fork in production, so the base of the code is working for us. As a stretch goal I'd like to implement a better storage structure to account for repeating elements (RLE).

@ddimaria I would still like to merge this since it would fix a lot of other feature requests. I would avoid the stretch goal for now.

src/xlsx/cells_reader.rs
```diff
@@ -134,6 +120,21 @@ where
...
+                    let style_id = if let Ok(Some(style_id_str)) =
+                        get_attribute(c_element.attributes(), QName(b"s"))
+                    {
+                        atoi_simd::parse::<usize>(style_id_str).unwrap_or(0)
+                    } else {
+                        0 // Default to style ID 0 when not present
+                    };
+
+                    if style_id < self.styles.len() {
+                        style = Some(self.styles[style_id].clone());
```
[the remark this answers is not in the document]
ddimaria: I'd like to get some perf number before implementing.  We've been using this branch in production and haven't noticed perf issues.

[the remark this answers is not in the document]
louisdewar: Sorry for the delay.

Simple benchmark:
```rs
use std::env;
use calamine::{open_workbook_auto, Reader};

fn main() {
    let file_path = env::args().nth(1).expect("Please provide a file path");

    let mut workbook = open_workbook_auto(&file_path)
        .expect("Failed to open workbook");

    for sheet_name in workbook.sheet_names() {
        let range = workbook
            .worksheet_range(&sheet_name)
            .expect("Failed to read worksheet");

        for row in range.rows() {
            for cell in row {
                std::hint::black_box(cell);
            }
        }
    }
}
```

## Before
```
# warmup = 2
Benchmark 1: ./target/release/examples/parse_tmp ~/Downloads/DESRI_BASE.xlsm
  Time (mean ± σ):      1.285 s ±  0.005 s    [User: 1.206 s, System: 0.075 s]
  Range (min … max):    1.275 s …  1.292 s    10 runs
```
https://share.firefox.dev/4i0pUAo
7.2% of total extraction time is spent doing a specific style.clone().

## After removing .clones
I'd imagine we would have comparible performance between this and just storing the style ID for each cell.
```
$ git diff
diff --git a/src/xlsx/cells_reader.rs b/src/xlsx/cells_reader.rs
index b1b1168..e60dc2f 100644
--- a/src/xlsx/cells_reader.rs
+++ b/src/xlsx/cells_reader.rs
@@ -132,7 +132,7 @@ where
                     };

                     if style_id < self.styles.len() {
-                        style = Some(self.styles[style_id].clone());
+                        // style = Some(self.styles[style_id].clone());
                     }

                     loop {
@@ -211,7 +211,7 @@ where
                     };

                     if style_id < self.styles.len() {
-                        style = Some(self.styles[style_id].clone());
+                        // style = Some(self.styles[style_id].clone());
                     }

                     loop {
@@ -352,7 +352,8 @@ where
                     {
                         if let Ok(style_id) = atoi_simd::parse::<usize>(style_id_str) {
                             if style_id < self.styles.len() {
-                                self.styles[style_id].clone()
+                                Style::new()
+                                // self.styles[style_id].clone()
                             } else {
                                 Style::new()
                             }
```

```
# warmup = 2
Benchmark 1: ./target/release/examples/parse_tmp ~/Downloads/DESRI_BASE.xlsm
  Time (mean ± σ):      1.103 s ±  0.005 s    [User: 1.038 s, System: 0.061 s]
  Range (min … max):    1.093 s …  1.110 s    10 runs
```
https://share.firefox.dev/47WkCBo

About a 14% reduction in time.

## Notes

I assume the easiest way to get this speedup is to simply store style ID in the cells and expose the style table.

I haven't profiled memory, I'm not sure what the best test of this is but I imagine this would also show a reasonable impact assuming you were storing the entire workbook in memory e.g. doing a .collect() on the cell reader.

Also this benchmark could easily be biased, I just picked a large xlsx I was using at work, other workbooks might show something different.

My laptop is quite fast (m4 pro), I can imagine the different is a lot bigger on slower servers etc...

jmcnamara:
@ddimaria Any update on this. I would really like to get this upstream in the next few months.

See also: #597

louisdewar:
I'm also willing to help out on this if that would be useful

jmcnamara:
@louisdewar that would be helpful.

ddimaria:
@jmcnamara I need to implement conditional formatting and validations, is it OK to add them to this PR?

jmcnamara:
@ddimaria This PR is already very large. It would be best to get this stabilized first (is it suitable for review now?). 

You can create other PRs that are dependent on this on. Conditional formatting and Data Validation should be separate PRs.

Erik1000:
> > @ddimaria This PR is already very large. It would be best to get this stabilized first (is it suitable for review now?).
> > You can create other PRs that are dependent on this on. Conditional formatting and Data Validation should be separate PRs.
> 
> That works for me. I just rebased my fork's commits onto upstream/master and squashed into a single commit. I'll do a final read through the code this morning and mark this ready for review.

is this PR now ready for review and can be merge? 👀

ddimaria:
> > > @ddimaria This PR is already very large. It would be best to get this stabilized first (is it suitable for review now?).
> > > You can create other PRs that are dependent on this on. Conditional formatting and Data Validation should be separate PRs.
> > 
> > 
> > That works for me. I just rebased my fork's commits onto upstream/master and squashed into a single commit. I'll do a final read through the code this morning and mark this ready for review.
> 
> is this PR now ready for review and can be merge? 👀

@Erik1000 I believe it is, though some conflicts just popped up.  Let me review those and CI and it will be ready to review/merge.

ddimaria:
@jmcnamara this PR is ready for review.  I'm staring conditional formatting in another branch.

jmcnamara:
> this PR is ready for review. I'm staring conditional formatting in another branch.

Ok. Fix the `cargo fmt` issues and if present, any clippy warnings. I think it would be best to resubmit this as a new PR. There are a lot of out of date comments on this PR.

Erik1000:
@ddimaria will you submit a new PR or is there already an existing one?

ddimaria:
I'm back on this today.  Will resolve conflicts and start a new PR

ddimaria:
PR moved to https://github.com/tafia/calamine/pull/653, it OK to close this one.

PR #539, opened 2025-07-25
Fix typos and add typos GH Actions workflow to prevent future typos

jqnatividad:
cc @tafia @jmcnamara @Expurple

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -1,0 +1,17 @@
+name: Typos Check
+on:
+  pull_request:
+  workflow_dispatch:
+  
+jobs:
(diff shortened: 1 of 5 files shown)


jmcnamara (changes requested):
It would be better to call this `.typos.toml` which is also allowed by the tool.

Also, I would have committed this with the CI change but that is up to you.

jmcnamara (changes requested):
It would be better to call this .typos.toml which is also allowed by the tool.

Also, I would have committed this with the CI change but that is up to you.

jmcnamara (approved):

PR #541, opened 2025-07-29
fix mut warning and fmt closing brace

azarmadr:


src/xls.rs
```diff
@@ -728,7 +728,7 @@ fn parse_mul_rk(
         });
     }
 
-    let mut col = col_first as u32;
```
jmcnamara, line 731: This does not look correct. `col` is clearly mutated below.

jmcnamara (changes requested):

src/xls.rs
azarmadr: I was experimenting with the code and added a todo! in the function call in the for loop. the compiler ingored the mutation and gave me the warning. It is incorrect change. i have remove the change. the other one is just a minor space update. Should i delete this pr?

jmcnamara: The other change is minor so I'll just fix that as part of some release fix-ups. 

I'll close this PR.

PR #542, opened 2025-08-02
unpin zip and using MSRV-aware in CI

Dirreke:
Detailed in #527

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,6 +14,9 @@ jobs:
           - nightly
     steps:
     - uses: actions/checkout@v4
+    - name: ensure dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -14,6 +14,9 @@ jobs:
           - nightly
     steps:
     - uses: actions/checkout@v4
+    - name: ensure dependencies
+      if: ${{ matrix.toolchain == '1.75' }}
```
jmcnamara, line 18: This would probably be better as:

```
if: ${{ matrix.toolchain < '1.85' }}
```

jmcnamara:
Overall this looks good. One minor comment inline.

PR #549, opened 2025-09-03
xls: fix issue with incorrect SST record

### LINKED ISSUE #548: Bug: Excel file will not parse
### Current behavior

A python repo using Calamine fails to parse a 'xls' file produced by Excel, see here: https://github.com/dimastbk/python-calamine/issues/150

Sorry, I don't know how to embed this in the Rust test stack.

### Expected behavior

The excel file should be parsed and the apparent user alerted to the error in the file.

### Sample code to reproduce

```rust
See downstream example
```

### Test file

[December-2022-AE-by-provider-UScya-revised-110523.xls](https://github.com/user-attachments/files/22120825/December-2022-AE-by-provider-UScya-revised-110523.xls)

### Environment

```text
- calamine version: unknown, but python-calamine   0.4.0
- Cargo.toml dependency line for calamine: n/a
- rustc version: unknown (see context)
- Excel/OpenOffice/LibreOffice version: unknown (downloaded published file from UK govt.)
- OS: OSX
```

### Checklist

- [ ] I have added a complete sample program that compiles in Rust.

jmcnamara:
This commit fixes an issue in the Excel xls SST record (Shared String Table) when the unique string count is incorrect.

Fixes #548


timcoote:
documentation point: the breakage is in a .xls file, not xlsx.

sftse:
If the `len` is generally considered untrusted, can we remove the `Vec::with_capacity(len)` above it? This kind of rightsizing seems like a footgun, see #529, and would potentially lead to OOM if we had a fuzzer sophisticated enough to reach those codepaths.

Will try to give a more thorough review within a day or two.

jmcnamara:
@timcoote 
> documentation point: the breakage is in a .xls file, not xlsx.

Thanks. Fixed.

> should the user be made aware of the error in the xls file?

In this particular case I don't think the information is useful to most end-users and unless they wrote the xls exporting software there isn't anything that they can do about it. If you know what software produced these files we could raise the issue upstream.

jmcnamara:
@timcoote 

> What's my fastest route for validating the fix against other files - build the dev environment, or wait for the PR to flow thro' ?

It depends on how urgently you need it. The fix should be released over the weekend at which point `python-calamine` can consume it.

If you want to test it in the interim you can modify the [`Cargo.toml file`](https://github.com/dimastbk/python-calamine/blob/master/Cargo.toml) for `python-calamine` and modify the `calamine` dependency entry to the following:

```
calamine = { 
    version = "0.30.0", 
    features = ["dates"] , 
    git = "https://github.com/jmcnamara/calamine.git", 
    branch = "gh548_incorrect_sst_unique_count"}
```

You will then need to rebuild the python module and install it. This will be slightly tricky if you haven't done it before.

timcoote:
Indeed. It failed on `pytest` - I'll chase up downstream. However, I also had to put the Cargo.toml dependency on one line for it to parse. Is this expected?

jmcnamara:
Merged so that the fix can be released to unblock downstream.

PR #551, opened 2025-09-05
cfb: replace built-in cfb handling with cfb crate

jmcnamara:


sftse:
One thing to consider is the deep dependency tree that `rust-cfb` acquired in order to fix quite an edge case of Unicode uppercasing [#62](https://github.com/mdsteele/rust-cfb/pull/62). I wasn't enthusiastic about it at the time, but maybe there's something that can be done before merging this change? That fix took `rust-cfb` from a lightweight dependency to quite an expensive one.

dimastbk:
Master: `test bench_xls          ... bench:  84,629,129.40 ns/iter (+/- 2,162,809.64)`
This PR: `test bench_xls          ... bench: 194,271,326.00 ns/iter (+/- 12,749,483.05)`

File https://github.com/MarkPflug/Benchmarks/blob/main/source/Benchmarks/Data/65K_Records_Data.xls

dimastbk:
> Did you use the Calamine "bench" code

Yes.

sftse:
I intend to investigate the performance regression if there is a consensus that this is necessary to merge, but can't say when I can get around to it.

jmcnamara:
I've updated this PR to pick up pick up cfb.rs v0.12.0 and rebased to main.

However, I am going to park/close this PR since the performance delta is too big. I would probably need to dig into the `cfg.rs` code to fix that.

sftse:
Since the 1. July 6 PRs from me were merged to the cfb crate, this seems to me a rather reasonable cadence for a project that has hardly any open issues. The most recent suggested fix took 10 days from PR to merge, again, rather reasonable. The release seems to have been forgotten, since a follow up request for a release was answered within 8h.

jmcnamara:
Update.

I noticed that one of the `cfb.rs` performance issues had been fixed (https://github.com/mdsteele/rust-cfb/issues/57) so I added the latest `cfb.rs` version (v0.12) and rebased this work to replace the internal cfb handling with the `cfb.rs` crate. However performance still looks to be ~2x slower.

I benchmarked with the current calamine criterion benchmarks but with the larger [65K_Records_Data.xls](https://github.com/MarkPflug/Benchmarks/blob/main/source/Benchmarks/Data/65K_Records_Data.xls) file suggest by @dimastbk above.

```diff
$ git diff
diff --git a/benches/basic.rs b/benches/basic.rs
index 82cbf01..267f822 100644
--- a/benches/basic.rs
+++ b/benches/basic.rs
@@ -61,7 +61,7 @@ fn count_cells_reader_xlsb(path: &str) -> usize {
 }
 
 fn bench_xls(c: &mut Criterion) {
-    c.bench_function("xls", |b| b.iter(|| count::<Xls<_>>("tests/issues.xls")));
+    c.bench_function("xls", |b| b.iter(|| count::<Xls<_>>("tests/65K_Records_Data.xls")));
 }
 
 fn bench_xlsx(c: &mut Criterion) {
```

Then running against the benchmark of the unaltered main/master:

```bash
$ git checkout master

$ cargo bench --bench basic -- --exact xls --save-baseline before

$ git checkout replace_cfb

$ cargo bench --bench basic -- --exact xls --baseline before
Benchmarking xls: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 29.5s, or reduce sample count to 10.
xls                     time:   [308.79 ms 315.51 ms 322.37 ms]
                        change: [+190.90% +197.44% +204.20%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

If someone has some time they can double check this and/or try to figure out where the bottleneck is in the updated `cfb.rs`. I would still prefer to use it for maintenance reasons but the performance would need to be a lot closer.

jmcnamara:
Reopening for visibility.

PR #552, opened 2025-09-06
fix: xlsx cell name handling for non-ASCII characters

hayandev:
Fix XLSX cell name handling to properly support non-ASCII characters by converting character-based processing to byte-based processing.


jmcnamara:
Thanks for the contribution. 

Since this is a "fix" PR is would be best to start with a  [Bug Report](https://github.com/tafia/calamine/issues/new?template=bug_report.yml) so that it is clear what is being fixed and so that a test can be added at the same time as the fix.

Could you please start there.

hayandev:
Thanks! I've created the related issue: #553

jmcnamara:
Thanks. 

Could you add the a test case like the following with the file from the bug report:

https://github.com/tafia/calamine/blob/master/tests/test.rs#L1776-L1790

Put a comment before the test case like:

```rust
// Test incrementing a shared formula in an xlsx file where the formula
// contains utf-8 characters.
```

If you know how then please do a `git commit --amend` with the new code and force push to your branch so that there is only one commit. 

Also change the commit message to say:

```
xlsx: fix shared formula handling with utf-8 characters

Closes #553 
```

jmcnamara:
Merged. Many thanks.

jmcnamara:
Upstream in v0.31.0. Thanks.

PR #556, opened 2025-09-16
deps: remove log.rs dependency

jmcnamara:
Removed log.rs dependency and debug statements. Replaced any required `warn!()` or `debug!()` with `eprintln!()`.

This cleans up the code and removed a dependency.

While refactoring/replacing the `cfb` code in #551 I noticed that the was a number of debug statements that weren't required or could be replaced by temporary `dgb!()` statements if needed.

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -8,8 +8,6 @@ use std::borrow::Cow;
 use std::cmp::min;
 use std::io::Read;
 
-use log::debug;
-
 use encoding_rs::{Encoding, UTF_16LE, UTF_8};
 
 use crate::utils::*;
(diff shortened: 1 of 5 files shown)


sftse:
I don't think this change should be done. `eprintln` are bad from a user and developer experience, and silencing warnings of that sort is only possible by closing stderr. Logging is quite small and is already pulled in as a transitive dependency.

KillTheMule:
As a user of calamine for a user-facing program, I heartily concur. I have gone through great lengths to make logging as useful as possible (my users are technically very non-experts), because that's the only real data point I have to diagnose problems. Writing to stderr is the same as not doing anything at all from this perspective, and valuable information might be lost.

PR #559, opened 2025-09-18
feat: reading PivotTable (PivotCache)

siqpush:
Feature to read all data available to a pivot table.

The data supporting a pivot table is referred to as the pivotCache. And I like to consider this feature "Calamine for your Cache".

An example use case would be auditing filtered content in an externally sourced pivot table.

Pivot Table's require both xl/pivotCache/PivotCacheDefinitions and xl/pivotCache/PivotCacheRecords files. The definitions file has relevant metadata as well as shared items. While the records file has values - rows are delimited with the \<r\> tag. \<x\>  indicates only their position in the Definitions file is given (sample below from tests/pivots.xlsx)

image

**xl/pivotCache/PivotCacheRecords1.xml**

```
<r>
    <n v="10"/>
    <s v="j"/>

    <!-- use data from PivotCacheDefinitions1.xml's 3rd 'CacheField Tag'  -->
    <x v="1"/>

    <x v="3"/>
    <n v="1.20452"/>
    <x v="5"/>
    <n v="4.1510311161292464"/>
    <b v="0"/>
    <m/>
    <s v="blue"/>
  </r>
```

**xl/pivotCache/PivotCacheDefinitions1.xml**

```
  <cacheFields count="10">
    <cacheField name="Id" numFmtId="0">
      <sharedItems containsSemiMixedTypes="0" containsString="0" containsNumber="1" containsInteger="1" minValue="1" maxValue="10"/>
    </cacheField>
    <cacheField name="Name" numFmtId="0">
      <sharedItems/>
    </cacheField>

    <!-- Corresponding lookup value <x v="1"/> above refers to <s v="blue"/> -->
    <cacheField name="Category" numFmtId="0">
      <sharedItems count="2">
        <s v="blue"/>
        <s v="yellow"/>
      </sharedItems>
    </cacheField>

    <cacheField name="Value" numFmtId="0">
      <sharedItems containsSemiMixedTypes="0" containsString="0" containsNumber="1" containsInteger="1" minValue="5" maxValue="20" count="4">
        <n v="10"/>
        <n v="20"/>
        <n v="15"/>
        <n v="5"/>
      </sharedItems>
```      

Example
```   


    let mut wb: Xlsx<_> = wb("pivots.xlsx");
    for result in wb.get_pivot_data_by_name_ref("PivotTable1").unwrap() {
        println!("{:?}", result);
    }

    /*
     prints the following:
     
    [String("Id"), String("Name"), String("Category"), String("Value"), String("Size"), String("Date"), String("Value / Size"), String("IsBlue"), String("Null"), String("Misc")]
    [Int(1), String("a"), String("blue"), Int(10), Float(1.78), DateTimeIso("2024-11-01T00:00:00"), Float(5.617977528089887), Bool(true), Empty, Empty]
    [Int(2), String("b"), String("blue"), Int(20), Float(2.012), DateTimeIso("2024-01-04T00:00:00"), Float(9.940357852882704), Bool(true), Empty, Float(2.012)]
...
*/
```

This may be determined to go outside the scope Calamine - but if it fits then it will need to be applied to other workbook formats (only .xlsx currently) and worked on in a few places like error handling.


jmcnamara:
Overall it looks okay. However there are a number of issues to fix before review:

- Rebase to master on calamine.
- Probably best to move the changes to a branch on your repo for easier  PRs.
- Turn on the CI on your branch and fix any issues before resubmitting the PR.
- Run `cargo fmt` on the code.
- Fix any `cargo clippy` issues.
- Fix the warnings from `cargo build -F pivot-cache`.
- Comment style should be proper sentence case with period at the end.
- Don't use `///` comments for non public comments. Use `//`.
- Explain why `fn xml_reader()` is being made public or don't make it public if it isn't required.
- Probably best to upgrade to Rust v1.89.0 or 1.90.0, if not already using them.
- Avoid making whitespace changes like removing blank lines unless there is a valid reason.

jmcnamara:
> but didn't realize it would automatically resubmit once I rebased it to my master. If that was the case I still have a little cleanup left removing the unnecessary comments.

That is fine, and normal. People often submit the PR as "draft" (it is an option in the initial GitHub dialog or you can do it in git) while they are iterating and then move it to full once it is ready to merge. 

> Also, do you plan to have this released for just .xlsx workbooks then have the remaining prd seperately?

I would think it would be too hard to do this for xls. It may be possible to do it for xlsb and I don't know about ods. So I think it is probably ok to work it out for just xlsx in this PR. 

Also, I will run the AI code review on this later. Use the usual amount of judgement in relation to the suggestions.

src/xlsx/mod.rs
```diff
@@ -711,6 +719,362 @@ impl<RS: Read + Seek> Xlsx<RS> {
...
+    // Unlike tables this is a direct opt-in and only needs to be called once to retrieve supporting information.
+    fn load_pivot_table_metadata(&mut self) -> Result<(), XlsxError> {
+        match self.read_pivot_table_metadata() {
+            Ok(pivot_tables) => {
+                if !pivot_tables.is_empty() {
+                    self.pivot_tables = Some(pivot_tables);
+                }
+                Ok(())
+            }
+            Err(e) => Err(e),
```
sftse, line 733: At a bare minimum we can use `?` here to handle the error path, but this also seems to contradict other functions in that it may lead to `pivot_tables = None` even if they were loaded. Other functions call an unwrap on these `pivot_tables`, like `fn pivot_tables`.

src/xlsx/mod.rs
```diff
@@ -711,6 +719,362 @@ impl<RS: Read + Seek> Xlsx<RS> {
...
+    ///
+    ///     // Check the pivot table names (ordering not garaunteed).
+    ///     assert_eq!(pivot_table_names, vec!["PivotTable1"]);
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    ///
+    pub fn pivot_tables_by_sheet(&self, sheet_name: &str) -> Result<Vec<&str>, XlsxError> {
+        if !self.sheet_names().contains(&sheet_name.to_string()) {
```
sftse, line 1029: This `to_string` can likely be removed.

src/pivot.rs
```diff
@@ -0,0 +1,186 @@
...
+        self.definitions = Some(s.into());
+        self
+    }
+
+    pub fn cache_number(mut self, n: usize) -> Self {
+        self.cache_number = Some(n);
+        self
+    }
+
+    pub fn build(self) -> PivotTableRef {
```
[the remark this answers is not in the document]
siqpush: I wish I could blame the AI but sadly no. The pattern seemed more useful in earlier iterations of the ```read_pivot_table_metadata``` function where I wasn't sure exactly how orderly I could get all the components - I think I had it in a map at some point. Left there because well I was probably trying to look cool and adhere to common patterns (which I guess is not dissimilar to what AI would do too).

src/xlsx/mod.rs
```diff
@@ -711,6 +719,362 @@ impl<RS: Read + Seek> Xlsx<RS> {
...
+            }
+            Err(e) => Err(e),
+        }
+    }
+    #[cfg(feature = "pivot-cache")]
+    // Finds pivot tables from sheet data and then finds their associated cache.
+    fn read_pivot_table_metadata(&mut self) -> Result<Vec<PivotTableRef>, XlsxError> {
+        let mut pivot_table_references = vec![];
+
+        for (i, (ref pivot_path, ref sheet)) in (self.find_pivot_table_paths_from_sheets()?)
```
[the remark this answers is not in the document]
siqpush: ref is dead to me ~ thanks for the informative link.

src/xlsx/mod.rs
```diff
@@ -1983,6 +2350,183 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+    pub struct PivotCache<'a, RS: Read + Seek + 'a> {
+        pub xl: &'a mut Xlsx<RS>,
+        pub cache_number: usize,
+    }
+
+    impl<'a, RS: Read + Seek + 'a> PivotCache<'a, RS> {
+        pub fn new(xl: &'a mut Xlsx<RS>, cache_number: usize) -> Self {
+            Self { xl, cache_number }
+        }
+    }
```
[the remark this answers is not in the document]
siqpush: See commit [1d646f4](https://github.com/tafia/calamine/pull/559/commits/1d646f43233a18c72ab1448e4eded6571bbeedb5) .

Apologies that the refactor makes the change a bit hard to track.

siqpush:
@jmcnamara I believe the failing check for stable appears to be due to the addition of clippy. The image installs the toolchain with `--profile minimal`. This setting does not include clippy by default. See line 48 in screen shot + commit which would conflict.

image


458b8ca7061b653e7be995976662a20003509986

jmcnamara:
@siqpush Thanks it is fixed on master. You will need to do a fetch and rebase to pick up the changes.

jmcnamara:
@siqpush Could you "Resolve" the comments that have been already addressed to make the review cleaner. Thanks.

jmcnamara:
I fixed the failing typo check on master. If you rebase to that it will fix the failing CI check.

src/xlsx/mod.rs
```diff
@@ -6,29 +6,32 @@
...
+use quick_xml::name::QName;
+use quick_xml::Reader as XmlReader;
+use std::borrow::Cow;
+use std::collections::BTreeMap;
+use std::io::BufReader;
+use std::io::{Read, Seek};
+use std::str::FromStr;
+use zip::read::{ZipArchive, ZipFile};
+use zip::result::ZipError;
+
```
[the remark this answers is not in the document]
jmcnamara: Unless I'm missing something it doesn't seem to have revert back to the way it was. In general it is a rule not to remove whitespace in a commit unless it relates to the commit. Having said that I could live with this change since the imports shouldn't have been separated out in the first place.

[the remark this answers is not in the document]
siqpush: No you are right my commit never landed in my master branch; but it should be fixed now. Probably better to just keep everything clean for the time being.

jmcnamara:
@siqpush I would like to merge this for the next release but there is a conflict due to recent changes. Could you fix this?

sftse:
@jmcnamara I don't think we should merge this in its current state. The commit history is very confusing (empty rebase commits, merge commits inside this PR) and I think the code could still use some improvements.
I can make another review.

jmcnamara:
> does commit [0b7a884](https://github.com/tafia/calamine/commit/0b7a88473fcd9541a3ee45058556a6802bac944e) better present the fix for you

Yes. The looks clean, from the point of view of a review but it seems to be detached from a branch.

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                                .rsplit_once('/')
+                                .expect("Must be a parent folder");
+                            paths.push(format!("{parent}/{target}"));
+                        } else if target.is_empty() { // do nothing
+                        } else {
+                            paths.push(target);
+                        }
+                    }
+                }
+                Ok(Event::End(ref e)) if e.local_name().as_ref() == b"Relationships" => break,
```
sftse: Ref can be removed, as previously mentioned
```suggestion
                Ok(Event::End(e)) if e.local_name().as_ref() == b"Relationships" => break,
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+        let mut xml = match super::xml_reader(zip, &rel_path) {
+            None => return Err(XlsxError::FileNotFound(rel_path.to_owned())),
+            Some(x) => x?,
+        };
+        let mut paths = vec![];
+        let mut buf = Vec::with_capacity(64);
+        loop {
+            buf.clear();
+            match xml.read_event_into(&mut buf) {
+                Ok(Event::Start(ref e)) if e.local_name().as_ref() == b"Relationship" => {
```
sftse:
```suggestion
                Ok(Event::Start(e)) if e.local_name().as_ref() == b"Relationship" => {
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+        {
+            let mut xml = match super::xml_reader(zip, &rel_path) {
+                // Some sheets may not have relationships - okay for path to not exist.
+                None => return Ok(vec![]),
+                Some(x) => x?,
+            };
+            loop {
+                buf.clear();
+                match xml.read_event_into(&mut buf) {
+                    Ok(Event::Start(ref e)) if e.local_name().as_ref() == b"Relationship" => {
```
sftse:
```suggestion
                    Ok(Event::Start(e)) if e.local_name().as_ref() == b"Relationship" => {
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                                    .rsplit_once('/')
+                                    .expect("Must be a parent folder");
+                                pivots_on_sheet.push(format!("{parent}/{target}"));
+                            } else if target.is_empty() { // do nothing
+                            } else {
+                                pivots_on_sheet.push(target);
+                            }
+                        }
+                    }
+                    Ok(Event::End(ref e)) if e.local_name().as_ref() == b"Relationships" => break,
```
sftse:
```suggestion
                    Ok(Event::End(e)) if e.local_name().as_ref() == b"Relationships" => break,
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+        let mut xml = match super::xml_reader(zip, pivot_path) {
+            None => return Err(XlsxError::FileNotFound(pivot_path.to_string())),
+            Some(x) => x?,
+        };
+        let mut buf = Vec::with_capacity(64);
+        let mut name = String::new();
+        loop {
+            buf.clear();
+            match xml.read_event_into(&mut buf) {
+                Ok(Event::Start(ref e)) if e.local_name().as_ref() == b"pivotTableDefinition" => {
```
sftse:
```suggestion
                Ok(Event::Start(e)) if e.local_name().as_ref() == b"pivotTableDefinition" => {
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                        if let Some(target) = target.strip_prefix("../") {
+                            // this is an incomplete implementation, but should be good enough for excel
+                            let (parent, _) = base_folder
+                                .rsplit_once('/')
+                                .expect("Must be a parent folder");
+                            paths.push(format!("{parent}/{target}"));
+                        } else if target.is_empty() { // do nothing
+                        } else {
+                            paths.push(target);
+                        }
```
sftse:
```suggestion
                        } else if !target.is_empty() {
                            paths.push(target);
                        }
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                    }
+                    if is_pivot_cache_record_type {
+                        if target.is_empty() {
+                        } else if target.starts_with("xl/pivotCache") {
+                            // do nothing
+                            paths.push(target);
+                        } else {
+                            paths.push(format!("xl/pivotCache/{target}"));
+                        }
```
sftse:
```suggestion
                        if target.starts_with("xl/pivotCache") {
                            paths.push(target);
                        } else if !target.is_empty() {
                            paths.push(format!("xl/pivotCache/{target}"));
                        }
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2435,474 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+    use quick_xml::name::QName;
+    use quick_xml::Decoder;
+    use std::collections::HashMap;
+    use std::io::{Read, Seek};
+
+    pub type Tag = Box<[u8]>;
+    pub type Value = Option<Box<[u8]>>;
+
+    // Get the target location of the pivot table's pivot cache definitions.
+    pub fn find_pivot_cache_definitions_from_pivot<RS>(
```
jmcnamara: Public functions should have full documentation (some of the functions elsewhere in this commit do and some don't.

jmcnamara: So should it be `pub(crate)` then?

tests/test.rs
```diff
@@ -2551,3 +2551,192 @@ fn test_xlsx_empty_shared_string() {
...
+    let separate_cache = pivot_tables
+        .iter()
+        .find(|pt| pt.name().eq("PivotTable5"))
+        .unwrap();
+    for pivot_table in pivot_tables.iter() {
+        if pivot_table.name().ne("PivotTable5") {
+            assert_ne!(pivot_table.definitions(), separate_cache.definitions())
+        }
+    }
+}
```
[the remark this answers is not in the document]
siqpush: Self serving implementation details. Watching out for non unique names at the workbook level, multiple pivots on a single sheet, cache spread over multiple pivot tables on multiple sheets...etc.

Long day with two sick little ones so a bit limited visualizing what a higher level would look like here. Do you have something in mind?

src/lib.rs
```diff
@@ -2302,3 +2302,65 @@ where
...
+/// Helper function to identify pivot tables by name and worksheet.
+///
+/// # Returns
+///
+/// ```text
+/// Vec<(String, String)>
+///        │       │
+///        │       └─── Worksheet name
+///        │
+///        └──── Pivot Table name
```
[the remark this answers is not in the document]
siqpush: I am too close to the code. I added this while I was navigating multiple pivot tables with the same name existing in a workbook. The unique identifier is the worksheet so this became an "attribute" of a pivot table. 

I doubt a user feels the same way so makes sense to add the change to next commit.

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+        Err(XlsxError::Unexpected(
+            "many definition cache relationships found for one pivot table",
+        ))
+    } else if paths.is_empty() {
+        Err(XlsxError::Unexpected(
+            "no cache definition found for pivot table",
+        ))
+    } else {
+        Ok(paths[0].clone())
+    }
```
sftse: We should be able to apply the same solution to the other similar code path we rewrote, but I was fond of the stronger validation where we check there is only a single path that could match: I think using an `Option` is the best course of action here, if we find a duplicate we error.

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                Ok(Event::End(e)) if e.local_name().as_ref() == b"Relationships" => break,
+                Ok(Event::Eof) => return Err(XlsxError::XmlEof("Relationships")),
+                Err(e) => return Err(XlsxError::Xml(e)),
+                _ => (),
+            }
+        }
+    }
+
+    Ok(pivots_on_sheet)
+}
```
sftse: This is my pattern-matching kicking in, is it correct that multiple pivots per sheet are permitted? Other functions error when more than one candidate item is found in the zip.

If you recheck, can you add a reference to the spec?

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                // This does not represent the underlying data and should be removed.
+                Ok(Event::Start(e)) if e.local_name().as_ref() == b"groupItems" => {
+                    field_names.pop();
+                    fields.pop();
+                }
+                Ok(Event::Start(e)) if item_tag(&e).is_some() => {
+                    if let Some(field) = fields.last_mut() {
+                        field.push((item_tag(&e).unwrap(), item_value(&e)?));
+                    }
+                }
```
sftse: Use pattern matching, any `is_some()/unwrap()` combo close by should be a candidate for elimination.

```suggestion
                Ok(Event::Start(e)) => {
                    if let Some(tag) = item_tag(&e) {
                        if let Some(field) = fields.last_mut() {
                            field.push((tag, item_value(&e)?));
                        }
                    }
                }
```

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+}
+
+impl PivotTables {
+    pub fn new() -> Self {
+        Self(vec![])
+    }
+
+    pub fn push(&mut self, pivot_table: PivotTableRef) {
+        self.0.push(pivot_table);
+    }
```
sftse: This function should be private.

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+}
+
+pub struct PivotTables(Vec<PivotTableRef>);
+
+impl Default for PivotTables {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
sftse: We can derive this `Default` impl.

sftse (commented):
Thanks for bringing this PR across the finish line, nearly there!

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+
+impl Deref for PivotTables {
+    type Target = [PivotTableRef];
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_slice()
+    }
+}
```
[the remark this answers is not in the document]
siqpush: https://github.com/tafia/calamine/pull/559/commits/6112834cbeb706a8540d55903cf2233d9e59ae28

removal of pub fn in this commit made reworking this no longer necessary so removed.

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                }
+                Ok(Event::Start(e)) if item_tag(&e).is_some() => {
+                    if let Ok(value) = item_value(&e) {
+                        row.push(parse_item(
+                            &(item_tag(&e).unwrap(), value),
+                            &self.reader.decoder(),
+                        ));
+                        col_number += 1;
+                    }
+                }
```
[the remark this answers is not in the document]
siqpush: https://github.com/tafia/calamine/pull/559/commits/a33dd97f90be8857dab497d009430c382a0c2b68

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+    };
+    let mut buf = Vec::with_capacity(64);
+    loop {
+        buf.clear();
+        match xml.read_event_into(&mut buf) {
+            Ok(Event::Start(e)) if e.local_name().as_ref() == b"Relationship" => {
+                let mut target = String::new();
+                let mut is_pivot_cache_record_type = false;
+                for a in e.attributes() {
+                    match a.map_err(XlsxError::XmlAttr)? {
```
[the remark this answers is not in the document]
siqpush: https://github.com/tafia/calamine/pull/559/commits/a33dd97f90be8857dab497d009430c382a0c2b68

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+            Some(x) => x?,
+        };
+        loop {
+            buf.clear();
+            match xml.read_event_into(&mut buf) {
+                Ok(Event::Start(e)) if e.local_name().as_ref() == b"Relationship" => {
+                    let mut target = String::new();
+                    let mut is_pivot_table_type = false;
+                    for a in e.attributes() {
+                        match a.map_err(XlsxError::XmlAttr)? {
```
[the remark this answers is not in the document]
siqpush: https://github.com/tafia/calamine/pull/559/commits/a33dd97f90be8857dab497d009430c382a0c2b68

src/xlsx/mod.rs
siqpush: https://github.com/tafia/calamine/pull/559/commits/1ae1490ff2b0e2fcdebd6076500bcf5a23c09e00

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+{
+    let mut pivots_on_sheet = vec![];
+    let mut buf = Vec::with_capacity(64);
+
+    let last_folder_index = sheet_path.rfind('/').expect("should be in a folder");
+    let (base_folder, file_name) = sheet_path.split_at(last_folder_index);
+    let rel_path = format!("{base_folder}/_rels{file_name}.rels");
+
+    // we need another mutable borrow of self.zip later so we enclose this borrow within braces
+    {
```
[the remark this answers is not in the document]
siqpush: No, no it is not.
https://github.com/tafia/calamine/pull/559/commits/c390ce599d87ea1ee8cb8fb09016f1dea1b78b92

src/xlsx/mod.rs
```diff
@@ -2199,6 +2280,685 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+    ///
+    /// }
+    /// ```
+    ///
+    pub fn get_pivot_tables_by_name_and_sheet(&self) -> Vec<(String, String)> {
+        self.0
+            .iter()
+            .map(|pt| (pt.sheet().to_string(), pt.name().to_string()))
+            .collect()
+    }
```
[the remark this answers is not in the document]
siqpush: I think it felt strange because I exposed functionality to `PivotTables` and even `PivotTableRef` that felt like they should be able to get data on their own. In turn this made functions like the above feel extra. 

As for the alternative, my gut thinking is that because we also have ```get_pivot_tables_by_sheet```, which would then also need be removed, we give the filter map to the user. Then at best case they are left having to do some extra doc reading on some of the uniqueness subtleties. 

As for a concrete example, I imagine for auditing data in workbook that might be sensitive to expose to the wrong user / client / ..etc. Using the relevant workbook from tests, this is how I would avoid leaking the top secret details of Category Blue:

```let path = "tests/pivots.xlsx";
let mut workbook: Xlsx<_> = open_workbook(path)?;
let pivot_tables = workbook.read_pivot_table_metadata()?;
for (sheet, pt) in pivot_tables.get_pivot_tables_by_name_and_sheet() {
    let mut check_col = 0;
    for (row_number, row) in workbook.pivot_table_data(&pivot_tables, sheet, pt)?.enumerate() {
        // header is the first row
        if row_number == 0 {
            for (col_number, field) in row.iter().enumerate() {
                if field.eq(&crate::calamine::Data::String("Category".to_string())) {
                    check_col = col_number
                }
            }
        } else if row[check_col].eq(&crate::calamine::Data::String("blue".to_string())) {
                panic!("Blue should not be included in this report.")
        }
    }
}```

File: tests/test.rs (hunk 15 above)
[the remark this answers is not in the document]
Comment (siqpush) [reply tests/test.rs]: see comment referencing below: https://github.com/tafia/calamine/pull/559#discussion_r2574317597

File: src/xlsx/mod.rs [hunk 28]
```diff
@@ -2199,6 +2284,676 @@ pub(crate) fn path_to_zip_path<RS: Read + Seek>(zip: &ZipArchive<RS>, path: &str
...
+                            key: QName(b"v"),
+                            value,
+                        }) = a
+                        {
+                            let value_position = self
+                                .reader
+                                .decoder()
+                                .decode(value.as_ref())
+                                .map(|val| val.parse::<usize>().unwrap())
+                                .unwrap();
```
sftse: These unwraps need be replaced with error handling.

siqpush: https://github.com/tafia/calamine/pull/559#discussion_r2608566995

PR #566, opened 2025-10-15
Fix dragged formula multiple axes

louisdewar:
Fixes #565

PR #568, opened 2025-10-15
Fix absolute shared formulas

### LINKED ISSUE #567: Bug: dragged formula with absolute references doesn't work
### Current behavior

Absolute references are not handled when formulas are dragged (shared formulas) - they remain constant.

=A$1 is dragged from A1 to E3

Incorrect output:
A$1;B$1
A$1;B$1
A$1;B$1

Running that code on my recent PR #566 gives:
A$1;B$1;B$1;B$1;B$1
A$1;B$1;B$1;B$1;B$1
A$1;B$1;B$1;B$1;B$1

which is closer but still wrong.

### Expected behavior

Absolute formulas are replaced:
A$1;B$1;C$1;D$1;E$1
A$1;B$1;C$1;D$1;E$1
A$1;B$1;C$1;D$1;E$1

### Sample code to reproduce

```rust
use std::env;
use std::path::PathBuf;

use calamine::{open_workbook_auto, Reader};

fn main() {
    let excel_file = env::args()
        .nth(1)
        .expect("Please provide an excel file to convert");

    let sheet_name = env::args()
        .nth(2)
        .expect("Expecting a sheet name as second argument");
    let excel_path = PathBuf::from(excel_file);

    let mut workbook = open_workbook_auto(&excel_path).unwrap();
    let range = workbook.worksheet_formula(&sheet_name).unwrap();
    let (end_row, end_column) = range.end().unwrap();
    for row in 0..=end_row {
        for column in 0..=end_column {
            print!(
                "{}",
                range.get_value((row, column)).map_or("(empty)", |v| v)
            );
            if column != end_column {
                print!(";")
            }
        }
        print!("\n")
    }
}
```

### Test file

[calamine_bug_2.xlsx](https://github.com/user-attachments/files/22932715/calamine_bug_2.xlsx)

### Environment

```text
- calamine version: latest from main
- Cargo.toml dependency line for calamine: N/A
- rustc version: 1.90.0
- Excel/OpenOffice/LibreOffice version: Excel on Mac Version 16.102 (25101223)
- OS: MacOS
```

### Checklist

louisdewar:
This PR includes changes from #566.

It should fix #567

### DIFF
--- a/src/xlsx/mod.rs
+++ b/src/xlsx/mod.rs
@@ -1951,24 +1951,26 @@ fn parse_cell_reference(name: &[u8]) -> Result<CellReference, XlsxError> {
 }
 
 // Format a cell reference (e.g., "$A$1", "A1", etc.).
-fn format_cell_reference(cell: &CellReference) -> Result<Vec<u8>, XlsxError> {
+fn format_cell_reference(cell: &CellReference, buf: &mut Vec<u8>) -> Result<(), XlsxError> {
+    if cell.col >= MAX_COLUMNS {
+        return Err(XlsxError::ColumnNumberOverflow);
+    }
     if cell.row >= MAX_ROWS {
         return Err(XlsxError::RowNumberOverflow);
     }
-    let mut buf = Vec::new();
     if cell.absolute_col {
         buf.push(b'$');
     }
-    buf.extend(column_number_to_name(cell.col)?);
+    column_number_to_name(cell.col, buf)?;
     if cell.absolute_row {
         buf.push(b'$');
     }
     buf.extend((cell.row + 1).to_string().into_bytes());
-    Ok(buf)
+    Ok(())
 }
 
 // Advance the cell name by the offset.
-fn offset_cell_name(name: &[u8], offset: (i64, i64)) -> Result<Vec<u8>, XlsxError> {
+fn offset_cell_name(name: &[u8], offset: (i64, i64), buf: &mut Vec<u8>) -> Result<(), XlsxError> {
     let cell = parse_cell_reference(name)?;
 
     let new_row = if cell.absolute_row {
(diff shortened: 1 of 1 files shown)


louisdewar:
Let me know if you think this approach is too ugly, I think there's probably a nicer way to do it but I can't figure one out that doesn't require some allocations or re-writing some of the other helper methods.

louisdewar:
@jmcnamara I've updated the code. My test xlsx also depends on "multi-axis shared formulas" so I still have those two commits in this branch and that #566 needs to be merged before this one.

The 2nd commit updates the code to avoid making a bunch of small allocations (I wrote the 1st commit in the same style as the existing code, and then the 2nd should have the same behaviour but avoid making those allocations).

louisdewar:
In terms of justifying the performance optimisation commit:

I got claude to generate a quick file in the examples dir which parses formulas and prints the time:
```
use calamine::{open_workbook_auto, Reader};
use std::env;
use std::time::Instant;

fn main() {
    let path = env::args()
        .nth(1)
        .expect("Usage: cargo run --example parse_formulas <path_to_excel_file>");

    let start = Instant::now();
    let mut workbook = open_workbook_auto(&path).expect("Failed to open workbook");
    let open_time = start.elapsed();

    let sheet_names = workbook.sheet_names();
    let mut total_formulas = 0;

    let parse_start = Instant::now();
    for sheet_name in &sheet_names {
        match workbook.worksheet_formula(sheet_name) {
            Ok(range) => {
                let count = range.used_cells().count();
                total_formulas += count;
                println!("Sheet '{}': {} formulas", sheet_name, count);
            }
            Err(e) => {
                eprintln!("Error reading formulas from sheet '{}': {:?}", sheet_name, e);
            }
        }
    }
    let parse_time = parse_start.elapsed();

    println!("\nTotal formulas parsed: {}", total_formulas);
    println!("Time to open workbook: {:?}", open_time);
    println!("Time to parse formulas: {:?}", parse_time);
    println!("Total time: {:?}", start.elapsed());
}
```

I then ran profiling before and after the optimisation commit on a large workbook (both were after the fix for absolute references):

[before](https://profiler.firefox.com/public/yw9jn8sj7czj4xzsd1vz8smm54ashbe0qa6qez0/flame-graph/?globalTrackOrder=0&profileName=parse_formulas%20%28before%20alloc%20optimisation%29%20%E2%80%93%20macOS%2015.6.1&symbolServer=http%3A%2F%2F127.0.0.1%3A3000%2F0ryfxgib9zw4l919d0r56qb4251f1h9h1rcbc6i&thread=0&v=12) ~1.4-1.5 seconds

[after](https://profiler.firefox.com/public/95b2avepbtx57xjaa3zxqxawcw93mx2bxd5mrs0/flame-graph/?globalTrackOrder=0&profileName=parse_formulas%20%28after%20alloc%20optimisation%29%20%E2%80%93%20macOS%2015.6.1&symbolServer=http%3A%2F%2F127.0.0.1%3A3000%2F3m0z0h7fl72dzyn4an0jij2dsjh00h65zdrkpyz&thread=0&v=12) ~1.2-1.3 seconds

In the flame graph section you can see that a lot of the time in `offset_cell_name` is spent doing allocations, and you can see that in the "after" one there are far fewer (although the ones that are left still take most of the time).

It's a small win, but my laptop is pretty fast and it's likely a bigger win on slower devices. Also this benchmarking was really rough and I only did a few repeats and only on one large workbook, but I think the flamegraph makes it clear that this is an improvement.

src/xlsx/mod.rs
```diff
@@ -1877,13 +1885,109 @@ where
...
+        cell.col
+    } else {
+        (cell.col as i64 + offset.1) as u32
+    };
+
+    format_cell_reference(&CellReference {
+        row: new_row,
+        col: new_col,
+        absolute_row: cell.absolute_row,
+        absolute_col: cell.absolute_col,
```
jmcnamara: This is a better approach. +1.

jmcnamara:
> In terms of justifying the performance optimisation commit:
> 
> I got claude to generate a quick file in the examples dir which parses formulas and prints the time:

That is good. Thanks for that.

Overall the PR looks good. Just a few minor comments.

jmcnamara:
Ok. Please rebase the fixup and it should be good to go.

PR #571, opened 2025-10-23
Accept XLUnicodeRichExtendedString without reserved tag

sftse:
Analogous to #525 where the same was done for `XLUnicodeString`, no file available for a test case.


jmcnamara (commented):
I think the change to `continue_record()` is good. 

However, I don't think the commit message does it justice. "refactor: remove distinction between None and vec.is_empty()" doesn't say where the distinction is being removed. For something like this I would add a commit message like:

```
xls: refactor CONTINUE record handling

Refactored Record::continue_record() to remove the distinction
between None and vec.is_empty().
```

sftse:
GTM

PR #572, opened 2025-10-24
fix: do best-effort decoding when codepage is unknown

sftse:
I've hit an xls file that has an invalid codepage of 0. Decoding reveals it to be ascii compatible, so could be anything.

```text
[2025-10-23T09:42:27Z DEBUG calamine::cfb] decoded as
utf16be: ç´æ‘¯æ±¥,
utf16le: ç‘³æ½¤æ•¬,
win1252: stdole
[2025-10-23T09:42:27Z DEBUG calamine::cfb] decoded as
utf16be: â©œä»ã€°ã€²ã€´ãŒ°â´°ã€°ã€­ã€°ã€°âµƒã€°ã€­ã€°ã€°ã€°ã€°ã€°ã¶ç´£ãˆ®ã€£ã€£äŒºå±—æ¥®æ‘¯ç³å±“ç¥³ååœ¶ã‘œç´æ‘¯æ±¥ãˆ®ç‘¬æˆ£ä½Œä” ä…µç‘¯æµ¡ç‘©æ½®,
utf16le: å°ªç­‡ã€°ãˆ°ã°ã€³ã€­ã€°â´°ã€°ã€°äŒ­ã€°â´°ã€°ã€°ã€°ã€°ã€°ã˜´â½â¸²âŒ°âŒ°ã©ƒåœæ¹©æ½¤ç·åœç¹ä½—ã™—å°´ç‘³æ½¤æ•¬â¸²æ±´â¢ä±â…ç•æ½´æ…­æ¥´æ¹¯,
win1252: *\G{00020430-0000-0000-C000-000000000046}#2.0#0#C:\Windows\SysWOW64\stdole2.tlb#OLE Automation
```

These strings are decoded inside `vba::set_lid`. We could either narrowly attempt to decode the strings inside just this function, since it does some validation on the decoded strings, looking for the presence of a `#` or b) use a global fallback decoding as this PR suggests and log that the codepage was invalid.

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -10,7 +10,7 @@ use std::io::Read;
 
 use log::debug;
 
-use encoding_rs::{Encoding, UTF_16LE, UTF_8};
+use encoding_rs::{Encoding, UTF_16LE, UTF_8, WINDOWS_1252};
 
 use crate::utils::*;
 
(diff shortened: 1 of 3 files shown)


sftse:
More generally, I don't think parsing of VBA is so important to justify failing to read the xls. We could also ignore the decoding issues for now and only parse vba on demand, circumventing the need to deal with it.

jmcnamara:
> More generally, I don't think parsing of VBA is so important to justify failing to read the xls. We could also ignore the decoding issues for now and only parse vba on demand, circumventing the need to deal with it.

Agreed.

PR #573, opened 2025-10-28
Fix error due to empty si tag in sharedStrings

jiaaro:
When the `xl/sharedStrings.xml` file contains `<si></si>` (rather than `<si><t></t></si>`) the document still opens properly in excel, but not in calamine. 

I've encountered this in the wild and our users sometimes upload files like this.

I created the test file by making a new document with two cells, then manually editing the `xl/sharedStrings.xml` to remove the `<t>` element for one of the values. I confirmed the test document still opens in Excel without errors.

Extra info: I'm not actually a Rust programmer, I'm using `python-calamine`. Maintainers should feel free to edit or re-implement this PR as they see fit. I won't take offense :)


jmcnamara:
I've pushed a PR to fix this to #576

PR #575, opened 2025-10-29
Oom alloc

### LINKED ISSUE #550: Bug: Out of memory error caused by over-allocation in cfb.rs Sectors::get_chain() parsing
### Current behavior

#529 added a testcase that triggers an OOM allocation. Currently Rust still treats allocator failures as aborts unless nightly features are used to override this behavior.

As a contributor to [rust-cfb](https://github.com/mdsteele/rust-cfb) that has run through TBs of data in production and never had an abort or panic, we might want to consider using it as a replacement to the current code.
Because CFB is used across .doc, .ppt, .xls and .msg files, it is likely it is better tested. It has ~5x the downloads of calamine.

### Expected behavior

No aborts and no panics.

### Sample code to reproduce

```rust
See #529
```

### Test file

See #529 

### Environment

```text

```

### Checklist

sftse:
New PR for #529. Fixes #550 

We can bound sector numbers based on the file size. If we try to access one beyond the end, we error. This prevents the OOM because the in memory buffer gets extended up to the offset of the file where the sector would presumably be. Large sector numbers translate to offsets, triggering the OOM.

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -36,6 +36,12 @@ pub enum CfbError {
     SectorBeyondEof(u32),
 }
 
+impl From<std::io::Error> for CfbError {
+    fn from(e: std::io::Error) -> CfbError {
+        CfbError::Io(e)
+    }
+}
+
 impl std::fmt::Display for CfbError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
@@ -179,7 +185,7 @@ enum SectorSize {
 impl Header {
     fn from_reader<R: Read>(f: &mut R) -> Result<(Header, Vec<u32>), CfbError> {
         let mut buf = [0u8; 512];
-        f.read_exact(&mut buf).map_err(CfbError::Io)?;
+        f.read_exact(&mut buf)?;
 
         // check ole signature
         let signature = buf
@@ -197,7 +203,7 @@ impl Header {
                 // sector size is 4096 bytes, but header is 512 bytes,
                 // so the remaining sector bytes have to be read
                 let mut buf_end = [0u8; 4096 - 512];
-                f.read_exact(&mut buf_end).map_err(CfbError::Io)?;
+                f.read_exact(&mut buf_end)?;
                 SectorSize::Large
             }
             s => {
@@ -270,7 +276,7 @@ impl Sectors {
             self.data.resize(end, 0);
             // read_exact or stop if EOF
             while len < end {
-                let read = r.read(&mut self.data[len..end]).map_err(CfbError::Io)?;
+                let read = r.read(&mut self.data[len..end])?;
                 if read == 0 {
                     return Ok(&self.data[start..len]);
                 }


src/cfb.rs
```diff
@@ -279,6 +279,7 @@ impl Sectors {
             Vec::new()
         };
         while sector_id != ENDOFCHAIN {
+            // FIXME: see tests/OOM_alloc3.xls, this call OOMs
```
jmcnamara: I think the changes in this commit should be merged with the next commit.

src/cfb.rs
```diff
@@ -78,7 +80,9 @@ impl Cfb {
     pub fn new<R: Read>(mut reader: &mut R, len: usize) -> Result<Cfb, CfbError> {
         // load header
         let (h, mut difat) = Header::from_reader(&mut reader)?;
-        let mut sectors = Sectors::new(h.sector_size, Vec::with_capacity(len));
+        assert_ne!(h.sector_size, 0);
+        let maxid = len / h.sector_size;
+        let mut sectors = Sectors::new(h.sector_size, Vec::with_capacity(len), maxid as u32);
```
jmcnamara: Wouldn't it be simpler to just use a fixed size for capacity like `Vec::with_capacity(1024)` here and elsewhere. I doubt that these are cases where over or under allocation of Vec will make any difference to performance.

src/cfb.rs
```diff
@@ -78,7 +80,9 @@ impl Cfb {
     pub fn new<R: Read>(mut reader: &mut R, len: usize) -> Result<Cfb, CfbError> {
         // load header
         let (h, mut difat) = Header::from_reader(&mut reader)?;
-        let mut sectors = Sectors::new(h.sector_size, Vec::with_capacity(len));
+        assert_ne!(h.sector_size, 0);
```
jmcnamara: Replace this assert with an error Result.

sftse: The intent was a bit unclear, I've cleaned that up. It can't be zero because of the validation done in `Header::from_reader`.

src/cfb.rs
jmcnamara: > The separate commit adding the test case is useful to verify the abort before verifying the fix.

In general it is best not to commit any changes that break the build process or the test suite since it makes running `git bisect` harder. I've worked on projects where both of these are strictly forbidden. So the test case should be rolled into the commit that fixes it.

tests/test.rs
```diff
@@ -17,12 +17,16 @@ use std::sync::Once;
 
 static INIT: Once = Once::new();
 
+fn test_path(name: &str) -> std::string::String {
```
jmcnamara: +1. This is a good change cleanup change.

I don't think the `env!("CARGO_MANIFEST_DIR")` is strictly required. I leave it out of some of the sample code. The test suite will definitely run without it. I don't know what cases, historically or otherwise, required it. Maybe you could consider dropping it here and just using `format!("tests/{name}")`.

tests/test.rs
```diff
@@ -2201,6 +2201,9 @@ fn test_oom_allocation() {
     let ws = xls.worksheets();
     assert_eq!(ws.len(), 1);
     assert_eq!(ws[0].0, "Colsale (Aug".to_string());
+
+    let path = test_path("OOM_alloc3.xls");
```
jmcnamara: Same comment as the higher level conversation: failing test cases should be merged with the commit that fixes them.

sftse:
In terms of merging policy, for me there are two independent axes: a) Should one rebase so that the resulting merge is 'trivial' and b) should the resulting merge commit be kept around. I'm strongly in favor of a) but don't have strong for/against on b), leaning slightly in favor of keeping the commits because the PRs then get referenced that caused the commits, discussion is easier to find.

sftse:
@jmcnamara Do you think we can merge without blocking on a decision on merge policy?

PR #578, opened 2025-10-30
chore: applied some minor clippy lint suggestions

jqnatividad:


jmcnamara:
> CONTRIBUTING.md with an accompanying clippy GH Action Workflow? WDYT?

That is a good idea. I've been meaning to something like that. I'll try draft something up.





> As to the clippy change with the inline comments - I find it useful as it makes it shorter, but I'm also OK omitting that change.

I'm fine with it either way.

PR #592, opened 2025-12-17
feat(xlsx): add pictures_with_positions method

mgcrea:
Hi!

Thanks for the great lib, I did need to properly extract images cell position info from a project I'm working on. Only works for xlsx files (either DrawingML anchors or 365 Rich Data). No breaking changes as I added a new `pictures_with_positions` method.


src/xlsx/mod.rs
```diff
@@ -249,9 +251,12 @@ pub struct Xlsx<RS> {
     is_1904: bool,
     /// Metadata
     metadata: Metadata,
-    /// Pictures
+    /// Pictures (raw, without position info)
     #[cfg(feature = "picture")]
     pictures: Option<Vec<(String, Vec<u8>)>>,
+    /// Pictures with position information
+    #[cfg(feature = "picture")]
+    pictures_with_positions: Option<Vec<Picture>>,
```
jmcnamara: Rather than an `Option<Vec<Picture>>` would it make sense to just have `Vec<Picture>`? If necessary the user can check if `pictures_with_positions` is empty.

src/xlsx/mod.rs
```diff
@@ -1491,6 +1988,8 @@ impl<RS: Read + Seek> Reader<RS> for Xlsx<RS> {
         xlsx.read_workbook(&relationships)?;
         #[cfg(feature = "picture")]
         xlsx.read_pictures()?;
+        #[cfg(feature = "picture")]
+        xlsx.read_pictures_with_positions()?;
```
jmcnamara: This approach causes the image related XML to be parsed twice: for `xlsx.read_pictures()` and `xlsx.read_pictures_with_positions()` which is inefficient. I get that you are trying to avoid breaking backward compatibility but I think it is the lesser of the two evils here. I'd suggest expanding the `xlsx.read_pictures()` function to add the new functionality.

src/xlsx/mod.rs
```diff
@@ -24,6 +24,8 @@ use crate::datatype::DataRef;
 use crate::formats::{builtin_format_by_id, detect_custom_number_format, CellFormat};
```
jmcnamara: Overall comment. Make sure to `rustfmt` the code additions/changes.

jmcnamara (changes requested):

jmcnamara:
@mgcrea Any update on this?

adamjgmiller:
Going to run a code review on this - hope i'm not misunderstanding the "needs work" and "help wanted" label - been looking for a good public PR to test a new code review tool on.

jmcnamara:
@adamjgmiller it doesn't need a code review. It needs help to enable the comments from the existing code review.

adamjgmiller:
### Code review

**Branch:** `feat-pictures_with_positions` → `master`
**PR:** #592 (open)
**Review ID:** `rev_01KRBXT2JYB7KJB7TYQB8JDMMK`
**Fix threshold:** not yet set (run `/adamsreview:fix [threshold]` to apply fixes)
**Sub-agent tokens:** 961,611 across 29 invocations

Found 34 findings across all lanes:
- Deep lane (correctness/security): 9 auto-fixable, 1 informational, 1 uncertain
- Light lane (ux/policy/architecture): 1 manual, 1 uncertain

---

## Deep lane — correctness & security

### ✓ Auto-fixable (9)

| # | Score | Impact | File | Issue |
|---|-------|--------|------|-------|
| F001 | 92 | correctness | `src/xlsx/mod.rs:1018-1026` | The second-pass `anchor_index` increments for every anchor end event regardless of whether the first pass pushed a Picture for that anchor, causing wrong coordinates to be written to wrong pictures when any anchor lacks a matching embed ID or media file. |
| F005 | 88 | correctness | `src/xlsx/mod.rs:1042-1045` | Rich Data pictures are only attempted if DrawingML yields zero pictures; a workbook containing both DrawingML-anchored pictures and Rich Data cell images will silently omit the Rich Data images. |
| F007 | 78 | correctness | `src/xlsx/mod.rs:747-813` | Drawing-loop only handles `twoCellAnchor` and `oneCellAnchor`; `absoluteAnchor` (EMU-pixel based, the third DrawingML anchor kind) is silently dropped, so pictures positioned with absolute coordinates are never returned by pictures_with_positions(). |
| F010 | 78 | correctness | `src/xlsx/mod.rs:911-925` | The `Event::Empty(b"blip")` branch is unreachable: `xml_reader` sets `config.expand_empty_elements = true`, so self-closing `<a:blip .../>` tags are emitted as Start+End pairs and only the `Event::Start` arm fires. Dead code that misleads readers about the parser's coverage. |
| F022 | 70 | security | `src/xlsx/mod.rs:719-734` | Both read_pictures and read_pictures_with_positions unconditionally load all xl/media bytes on every workbook open, doubling peak memory; an attacker-supplied xlsx with large embedded images can double the heap allocation compared to pre-PR behaviour. |
| F023 | 78 | security | `src/xlsx/mod.rs:930-942` | For each drawing anchor that references the same media file, data.clone() produces a full copy of the image bytes; an attacker embedding one large image referenced by thousands of twoCellAnchor elements across sheets can force unbounded memory growth proportional to anchor_count × image_size. |
| F029 | 60 | correctness | `src/xlsx/mod.rs:685-1050` | pictures() returns every file under xl/media/, but pictures_with_positions() only returns images reachable via a sheet→drawing→rels chain; orphaned media (images not referenced by any drawing, or referenced only via header/footer/VML) are dropped in the new API. Callers picking the new API silently lose pictures the old API returned. |
| F032 | 78 | correctness | `src/xlsx/mod.rs:1147-1196` | The rich-data path treats any cell <c> with a vm attribute as a rich-value image, but vm is a generic 'value metadata' indirection through xl/metadata.xml that can reference other metadata kinds (dynamic-array spill, future feature blocks). The code skips xl/metadata.xml entirely and uses vm directly as a 1-indexed richValueRel index — producing false positives or wrong picture-cell mappings on workbooks with non-image vm usage. |
| F033 | 60 | correctness | `src/xlsx/mod.rs:892-925` | blip embed_id extraction is gated on in_anchor but not on being inside a pic/blipFill — any <blip> element anywhere in an anchor (including inside <a:extLst> or alternate-content fallbacks) overwrites current_embed_id, so an anchor with multiple blips or non-pic blip references can end up associated with the wrong media file. |

**Cross-cutting group G1:** F001 + F007 + F010 + F033 — Rewrite read_pictures_with_positions' DrawingML parser as a single pass over the events. The new pass tracks in_anchor, in_blipFill, in_pic, in_row, in_col state and an Event::Text handler accumulates row/col digits; at anchor end (twoCellAnchor, oneCellAnchor, absoluteAnchor) it pushes one Picture iff blip+embed+media resolved, otherwise pushes nothing — eliminating the two-pass anchor_index drift. Add `| b"absoluteAnchor"` to the anchor start/end match arms (F007). Delete the Event::Empty(b"blip") arm — unreachable under expand_empty_elements=true (F010). Gate current_embed_id capture on in_anchor && in_blipFill && in_pic so non-canonical or AlternateContent blips can't clobber the active embed (F033). Doing these in one rewrite avoids overlapping match-arm edits and ensures the new single-pass invariants hold for all three anchor kinds with correctly scoped blip capture.
**Cross-cutting group G2:** F022 + F023 — Single xl/media scan that produces an Arc-shared backing store. Change media_map to HashMap<String, Arc<[u8]>>, scan xl/media/ once in read_pictures, populate self.pictures with Arc<[u8]>, and have read_pictures_with_positions reuse self.pictures (or the shared map) instead of re-scanning the zip. Picture.data becomes Arc<[u8]> so per-anchor association is Arc::clone, not bytes copy. This is one coordinated change — keeping the second scan while flipping to Arc would still double-load the bytes, and switching to Arc without consolidating the scans leaves the duplicate-load DoS surface intact.
**Cross-cutting group G3:** F005 + F029 — Restructure the tail of read_pictures_with_positions in a fixed order: (1) remove the `if pictures.is_empty()` guard so read_rich_data_pictures runs unconditionally and appends rich-data entries; (2) AFTER rich-data completes, walk media_map and append orphan-media entries for any media file not already represented in pictures (anchor_row/anchor_col/sheet_name = None). Order matters: appending orphans before rich-data would cause rich-data images to be double-counted as orphans. Both changes mutate the same trailing section and must be designed together to keep the dedup set consistent.

<details><summary>Details and fix proposals</summary>

#### F001 — The second-pass `anchor_index` increments for every anchor end event regardless of whether the first pass pushed a Picture for that anchor, causing wrong coordinates to be written to wrong pictures when any anchor lacks a matching embed ID or media file.

**File:** `src/xlsx/mod.rs:1018-1026`
**Score:** 92 (strong)
**Reason:** Confirmed: Pass 1 only pushes a Picture when blip+embed+media resolve; Pass 2 increments anchor_index unconditionally per anchor end event. Any non-image anchor (chart graphicFrame, group shape, unsupported media) causes all subsequent pictures on the sheet to receive wrong row/col. Fix: collapse to single-pass parse (add in_row/in_col + Event::Text handler in Pass 1, push at anchor end, drop Pass 2).

#### F005 — Rich Data pictures are only attempted if DrawingML yields zero pictures; a workbook containing both DrawingML-anchored pictures and Rich Data cell images will silently omit the Rich Data images.

**File:** `src/xlsx/mod.rs:1042-1045`
**Score:** 88 (strong)
**Reason:** Confirmed: `if pictures.is_empty()` at line 1043 unconditionally skips read_rich_data_pictures whenever any DrawingML picture exists. read_rich_data_pictures is strictly additive (uses pictures.push) and gracefully early-returns when richValueRel.xml is absent, so calling unconditionally is safe. Fix: remove the is_empty guard. Add a regression test for mixed-format workbooks.

#### F007 — Drawing-loop only handles `twoCellAnchor` and `oneCellAnchor`; `absoluteAnchor` (EMU-pixel based, the third DrawingML anchor kind) is silently dropped, so pictures positioned with absolute coordinates are never returned by pictures_with_positions().

**File:** `src/xlsx/mod.rs:747-813`
**Score:** 78 (strong)
**Reason:** Confirmed empirically: a synthesized absoluteAnchor xlsx (built from picture.xlsx) drops to 1 picture vs baseline 2. Fix: add `| b"absoluteAnchor"` to the four match arms (lines 883/929/993/1018). absoluteAnchor without blip (chart-only) stays harmless. EMU pos coordinates remain unsurfaced (anchor_row/anchor_col=None) — separate feature gap.

#### F010 — The `Event::Empty(b"blip")` branch is unreachable: `xml_reader` sets `config.expand_empty_elements = true`, so self-closing `<a:blip .../>` tags are emitted as Start+End pairs and only the `Event::Start` arm fires. Dead code that misleads readers about the parser's coverage.

**File:** `src/xlsx/mod.rs:911-925`
**Score:** 78 (strong)
**Reason:** Confirmed dead code: xml_reader sets config.expand_empty_elements=true unconditionally (line 2152-2158), so self-closing `<a:blip/>` is emitted as Start+End pairs only. The Event::Empty(b"blip") arm at lines 911-925 is unreachable. The Event::Start arm at lines 892-907 handles blip identically. Removal is purely dead-code cleanup; the outer `_ => ()` catchall at line 960 makes it mechanically safe.

#### F022 — Both read_pictures and read_pictures_with_positions unconditionally load all xl/media bytes on every workbook open, doubling peak memory; an attacker-supplied xlsx with large embedded images can double the heap allocation compared to pre-PR behaviour.

**File:** `src/xlsx/mod.rs:719-734`
**Score:** 70 (moderate)
**Reason:** Confirmed: both read_pictures (685-710) and read_pictures_with_positions (715-739) independently iterate xl/media/ and call read_to_end on every entry. Both invoked unconditionally at workbook open (lines 1990, 1992). Peak heap is 2x media bytes (3x transiently when also cloned into Picture.data). Fix: build a shared media_index in read_pictures, have read_pictures_with_positions reuse self.pictures rather than scan the zip again.

#### F023 — For each drawing anchor that references the same media file, data.clone() produces a full copy of the image bytes; an attacker embedding one large image referenced by thousands of twoCellAnchor elements across sheets can force unbounded memory growth proportional to anchor_count × image_size.

**File:** `src/xlsx/mod.rs:930-942`
**Score:** 78 (strong)
**Reason:** Confirmed memory amplification: data.clone() at line 936 (and parallel 1181) produces full bytes copy per anchor. xlsx places no spec limit on anchor count and one media file can legally be referenced by unbounded anchors. A crafted xlsx (a few hundred KB on disk) with N anchors referencing one large image expands to N × image_size in resident memory. DoS amplification for parsers ingesting untrusted xlsx. Fix: change media_map to Arc<[u8]>, Picture.data to Arc<[u8]>, replace .clone() with Arc::clone (breaking API change).

#### F029 — pictures() returns every file under xl/media/, but pictures_with_positions() only returns images reachable via a sheet→drawing→rels chain; orphaned media (images not referenced by any drawing, or referenced only via header/footer/VML) are dropped in the new API. Callers picking the new API silently lose pictures the old API returned.

**File:** `src/xlsx/mod.rs:685-1050`
**Score:** 60 (moderate)
**Reason:** Confirmed by empirical repro: pictures() returns 3 entries vs pictures_with_positions() returns 2 when an orphan image is added to xl/media/. Header/footer images (sheet rels Type .../image or .../vmlDrawing) and orphaned media silently dropped from the new API. Picture struct already declares sheet_name/anchor_row/anchor_col as Option, supporting unanchored entries — but the code never emits them. Fix: at end of read_pictures_with_positions, walk media_map and append unanchored entries for media not yet present in pictures.

#### F032 — The rich-data path treats any cell <c> with a vm attribute as a rich-value image, but vm is a generic 'value metadata' indirection through xl/metadata.xml that can reference other metadata kinds (dynamic-array spill, future feature blocks). The code skips xl/metadata.xml entirely and uses vm directly as a 1-indexed richValueRel index — producing false positives or wrong picture-cell mappings on workbooks with non-image vm usage.

**File:** `src/xlsx/mod.rs:1147-1196`
**Score:** 78 (strong)
**Reason:** Confirmed: code uses vm directly as 1-indexed richValueRel index, skipping the xl/metadata.xml indirection. OOXML defines vm as a valueMetadata index whose <bk> entries can be XLRICHVALUE OR XLDAPR (dynamic-array spill) OR other kinds. Excel 365 workbooks combining =IMAGE() with =SEQUENCE/=SORT will misroute. Fix: parse xl/metadata.xml first to build vm-to-rvb-index map; only XLRICHVALUE bks resolve to image rvb indices. Self-contained to read_rich_data_pictures.

#### F033 — blip embed_id extraction is gated on in_anchor but not on being inside a pic/blipFill — any <blip> element anywhere in an anchor (including inside <a:extLst> or alternate-content fallbacks) overwrites current_embed_id, so an anchor with multiple blips or non-pic blip references can end up associated with the wrong media file.

**File:** `src/xlsx/mod.rs:892-925`
**Score:** 60 (moderate)
**Reason:** Confirmed latent bug: blip embed extraction gates only on in_anchor; any <a:blip> overwrites current_embed_id. OOXML legitimately allows multiple blips per anchor (AlternateContent fallbacks, grpSp groups, sp with blipFill). tests/picture.xlsx happens to have one blip per anchor so the bug is latent. Fix: track in_blipFill and in_pic flags; gate blip capture on `in_anchor && in_blipFill && in_pic` (canonical <xdr:pic><a:blipFill><a:blip> path only).

</details>

### Auto-recommendations (2)

_AI-authored fix directions for findings that aren't auto-fixable today. Run `/adamsreview:fix` to batch-apply with one confirmation, or `/adamsreview:walkthrough` to review one-by-one._

| # | Score | Disp | File | Recommendation |
|---|-------|------|------|----------------|
| F041 | 62 | manual | `src/lib.rs:444-449` | _high:_ Implement `pictures_with_positions` on `Xls`, `Xlsb`, and `Ods` (in `src/xls.rs`, `src/xlsb/mod.rs`, `src/ods.rs`) by forwarding to each ... |
| F043 | 60 | informational | `tests/test.rs:1519-1576` | _medium:_ Extend `tests/test.rs` with at least three additional `pictures_with_positions` sub-cases: (1) a multi-anchor fixture (two or more images... |

<details><summary>Full recommendations and alternatives</summary>

#### F041 — The trait default impl returns None for non-Xlsx formats, but auto.rs explicitly dispatches to e.pictures_with_positions() for Xls/Xlsb/Ods — all of which inherit the trait default. This means non-Xlsx readers silently return None even when pictures() returns image bytes, with no documented indication. Users iterating workbooks via Sheets enum get inconsistent behavior depending on file type.

**File:** `src/lib.rs:444-449`
**Score:** 62 (moderate)
**Reason:** Confirmed: auto.rs dispatches pictures_with_positions() to all four formats (Xlsx, Xls, Xlsb, Ods); Xls/Xlsb/Ods inherit trait default returning None while pictures() returns data for those formats. Inconsistency is undocumented. Fix options: (a) document the Xlsx-only behavior in the trait doc; (b) implement pictures_with_positions for Xls/Xlsb/Ods (with anchor_row/col=None where positions not extracted).
**Auto-recommendation (high):** Implement `pictures_with_positions` on `Xls`, `Xlsb`, and `Ods` (in `src/xls.rs`, `src/xlsb/mod.rs`, `src/ods.rs`) by forwarding to each format's existing `pictures()` and wrapping results into `Picture` structs with `sheet_name`, `anchor_col`, and `anchor_row` set to `None` — indicating position data is unavailable rather than silently returning `None` as if no pictures exist at all. If full position support is out of scope, at minimum update the trait-level doc-comment on `pictures_with_positions` in `src/lib.rs` to explicitly state which formats support position data and which return `None`, and add a note in the `Sheets::pictures_with_positions` dispatch in `src/auto.rs`.
- _Alternatives:_ **B** Doc-only mitigation

#### F043 — Both new tests use single-anchor drawings (picture.xlsx: 1 anchor per drawing; picture_richdata.xlsx: 1 rich-data image). Neither exercises (a) drawings with multiple anchors where the index-misalignment bug would surface, (b) absoluteAnchor handling, (c) workbooks mixing DrawingML and rich-data, (d) sheets with orphaned media in xl/media not referenced by any drawing, (e) anchors with row OR col text parse failure. The test surface is much narrower than the implementation surface, masking the bugs above.

**File:** `tests/test.rs:1519-1576`
**Score:** 60 (moderate)
**Reason:** Confirmed test coverage gap: both new tests use single-anchor fixtures. Recommended new fixtures: (a) picture_multi_anchor.xlsx (3 anchors, 3 media — exercises F001 index drift); (b) picture_absolute_anchor.xlsx (absoluteAnchor — exercises F007); (c) picture_mixed_drawingml_richdata.xlsx (exercises F005); (d) picture_orphaned_media.xlsx (exercises F029); (e) picture_partial_anchor.xlsx (malformed row/col text — exercises silent .parse().ok() partial-position class). Pair with the corresponding parser fixes.
**Auto-recommendation (medium):** Extend `tests/test.rs` with at least three additional `pictures_with_positions` sub-cases: (1) a multi-anchor fixture (two or more images on the same sheet) asserting correct per-anchor `anchor_col`/`anchor_row` values to exercise the index-alignment path; (2) a negative-path case loading a workbook where one anchor XML contains a non-numeric row or col value, asserting the result is either `None` or omits that anchor rather than panicking; (3) a mixed workbook (if a suitable fixture exists or can be constructed) with both DrawingML and richData images to confirm both paths populate the same returned `Vec<Picture>`. The fixtures can be minimal hand-crafted xlsx files committed to `tests/` alongside the existing `picture.xlsx`.
- _Alternatives:_ **B** Parameterized table-driven approach

</details>

### ℹ Uncertain (1)

| # | Score | Impact | File | Issue |
|---|-------|--------|------|-------|
| F025 | 45 | security | `src/lib.rs:188-191` | anchor_cell() computes row + 1 on a u32 parsed directly from attacker-controlled XML text with no bounds check; if an xlsx provides row value 4294967295 (u32::MAX), release builds silently wrap to 0 producing an incorrect cell reference, and debug builds panic. |

Phase 4 couldn't confirm decisively. Re-run `/adamsreview:review` if you suspect this deserves
further investigation with fresh context.

### ℹ Confirmed — informational (1)

| # | Score | Impact | File | Issue |
|---|-------|--------|------|-------|
| F043 | 60 | correctness | `tests/test.rs:1519-1576` | Both new tests use single-anchor drawings (picture.xlsx: 1 anchor per drawing; picture_richdata.xlsx: 1 rich-data image). Neither exercises (a) drawings with multiple anchors where the index-misalignment bug would surface, (b) absoluteAnchor handling, (c) workbooks mixing DrawingML and rich-data, (d) sheets with orphaned media in xl/media not referenced by any drawing, (e) anchors with row OR col text parse failure. The test surface is much narrower than the implementation surface, masking the bugs above. |

## Light lane — ux, policy, architecture

_Light-lane findings — including rows labeled auto-fixable — aren't applied by `/adamsreview:fix` directly. Findings with an auto-recommendation get batch-confirmed at `:fix`'s Phase 7.5 preflight (or `:walkthrough` Step 4.5); use `/adamsreview:promote <finding_id>` for a single-finding manual override._

| # | Score | Impact | File | Finding | Disposition |
|---|-------|--------|------|---------|-------------|
| F041 | 62 | ux | `src/lib.rs:444-449` | The trait default impl returns None for non-Xlsx formats, but auto.rs explicitly dispatches to e.pictures_with_positions() for Xls/Xlsb/Ods — all of which inherit the trait default. This means non-Xlsx readers silently return None even when pictures() returns image bytes, with no documented indication. Users iterating workbooks via Sheets enum get inconsistent behavior depending on file type. | manual |
| F014 | 57 | architecture | `src/xlsx/mod.rs:712-1051` | `read_pictures_with_positions` runs eagerly during workbook open (called unconditionally from the Reader impl, line 1992) for every xlsx file even when the consumer never asks for positions. It clones every media buffer per picture (`data.clone()`) and walks all sheet/drawing XML, materially increasing open cost and memory for the picture feature. Consider lazy population (parse-on-demand) and a single Arc<Vec<u8>> for shared media data. | uncertain |

## Polish — below threshold, clustered (11)

Below-gate findings (score < 45) that cluster in the same area — not worth surfacing individually, but dense enough that a human pass may catch something the pipeline filtered out.

| # | Score | File | Rough location | Concern |
|---|-------|------|----------------|---------|
| F018 | 25 | `src/xlsx/mod.rs` | L727-743 | Media-map step keys by raw `filename` only. If two different drawing rels target media via paths that share a basename (e.g. `xl/media/image1.png` and `xl/media/subfolder/image1.png`, allowed by the OPC spec), the second overwrites the first in `media_map`. The drawing-rels parser also keeps only the basename via `target.rsplit('/').next()`, so name collisions yield wrong picture content. |
| F012 | 20 | `src/xlsx/mod.rs` | L747-752 | Sheet rels path is built from `rfind('/').unwrap_or(0)`. When `sheet_path` contains no `/` (defensive but possible if sheets registry holds an unqualified name), `base_folder` becomes empty and `sheet_rels_path` is `"/_rels/<name>.rels"` with a leading slash — which will not match any zip entry. Same shape bug at `drawing_rels_path` computation (lines 819-823). |
| F038 | 35 | `src/xlsx/mod.rs` | L748 | `&self.sheets.clone()` is called twice (once in read_pictures_with_positions outer loop, once in read_rich_data_pictures step 3). Each clone duplicates the entire (name, path) vector. For large workbooks with hundreds of sheets, this is unnecessary allocation, and it exists only to dodge the borrow checker — the inner loop holds &mut self.zip while iterating the sheets vec. |
| F050 | 35 | `src/xlsx/mod.rs` | L754 | The sheet drawing is selected by scanning .rels for the first relationship whose type ends with /drawing rather than reading the worksheet's actual <drawing r:id="..."> element, so stale or extra relationships can cause wrong drawings to be used. |
| F051 | 25 | `src/xlsx/mod.rs` | L757 | New relationship and Rich Data parsers break on Event::Eof without verifying the closing root element, so truncated .rels or richValueRel.xml files silently produce partial picture metadata with no parse error. |
| F040 | 20 | `src/xlsx/mod.rs` | L778-781 | Drawing Relationship Type is matched by `value.ends_with(b"/drawing")`. The canonical relationship type is `http://schemas.openxmlformats.org/officeDocument/2006/relationships/drawing`, which ends with `/drawing` — but other relationship types like `.../vmlDrawing` end with `Drawing` (different casing) and don't collide, while custom or strict-OOXML namespaces are NOT relationship-type URIs but URIs in general — this is fragile if Microsoft introduces a new relationship type ending in `/drawing` (e.g., for chart sub-drawings). |
| F013 | 30 | `src/xlsx/mod.rs` | L791-800 | Relative-target resolution hardcodes the workbook root as `xl/`. `target.starts_with("../")` is rewritten as `format!("xl/{...}")`, which is wrong if the workbook was packaged with a non-default part path (e.g., custom Office Open XML packaging). Should resolve against the actual base of `sheet_path` rather than a hardcoded literal. |
| F019 | 10 | `src/xlsx/mod.rs` | L875-895 | `b"row"`/`b"col"` gating uses `in_from` but the drawing schema also contains `<xdr:from><xdr:col>...<xdr:row>...</></>` and `<xdr:to><xdr:col>...<xdr:row>...</></>`. In Pass 2 the gate is correct (`in_row if in_from`), but inside `<from>` there is also `<colOff>` and `<rowOff>` (EMU offsets). The matcher `b"row" if in_from` and `b"col" if in_from` will NOT match `b"colOff"`/`b"rowOff"` (different local names), so this is safe — but flagging because the parallel code path in Pass 1 never tracks these and the asymmetry should be examined. |
| F042 | 20 | `src/xlsx/mod.rs` | L1109-1137 | richValueRel.xml parsing scans for ANY <rel> element with an id attribute, with no parent-element gate. The order of <rel> elements observed becomes the rid_order index — but element ordering is determined by document order in the XML, which is well-defined per the OOXML spec but not validated. If a future revision interleaves <rel> elements with other content, order may not map 1:1 to vm indices. |
| F047 | 40 | `src/xlsx/mod.rs` | L1150 | Rich Data image cells are silently dropped when the <c> element has no explicit r attribute because the parallel parser does not track row/column position the way the main cell reader does. |
| F048 | 30 | `src/xlsx/mod.rs` | L1164 | saturating_sub(1) on vm_index silently maps an invalid vm="0" to index 0, ambiguously aliasing the first image and making every image off by one if the producer uses zero-based metadata. |

---

🤖 Generated with Adam's Claude Code Review Command

adamjgmiller:
> @adamjgmiller it doesn't need a code review. It needs help to enable the comments from the existing code review.

I didn't see the existing code review, sorry, and I already ran it! Do you mind if I leave the comment up here? I was looking for public PRs needing review as examples for my review plugin I built: https://github.com/adamjgmiller/adamsreview

Is there anything useful about the review afterall? Would be helpful for me to know.

adamjgmiller:
and would you like me to use my tool to fix any of the findings? least i can do after posting this!

jmcnamara:
@adamjgmiller 

> Is there anything useful about the review afterall? Would be helpful for me to know.

No. You can delete it. It just adds noise to an already existing code review. In fact it creates a secondary submission that needs to be reviewed so overall it juat adds to the work.

I don't mind existing maintainers or submitters using agents to support their code reviews but a standalone agentic review doesn't add any value if there isn't an associated human effort to evaluate it and make the changes required.

I don't think this type of submission would be a welcome addition to any open source project that I work on.

adamjgmiller:
> @adamjgmiller
> 
> > Is there anything useful about the review afterall? Would be helpful for me to know.
> 
> No. You can delete it. It just adds noise to an already existing code review. In fact it creates a secondary submission that needs to be reviewed so overall it juat adds to the work.
> 
> I don't mind existing maintainers or submitters using agents to support their code reviews but a standalone agentic review doesn't add any value if there isn't an associated human effort to evaluate it and make the changes required.
> 
> I don't think this type of submission would be a welcome addition to any open source project that I work on.

It's gone. Sorry to have bothered you!

jmcnamara:
Moved to #652.

PR #596, opened 2026-01-07
ods: fix DoS via repeated rows/columns (issue #594)

### LINKED ISSUE #594: Bug: DoS via repeated rows/columns in ODS files
### Current behavior

Parsing a malformed 7KB ODS file causes ~666ms processing time due to expansion of repeated rows/columns. The file claims 1M+ rows × 16K columns = 17+ billion cells, which calamine attempts to materialize in memory.

## Root Cause Analysis

The ODS format allows compact representation of empty cells via repeat attributes:
```xml
<table:table-row table:number-rows-repeated="1048576">
  <table:table-cell table:number-columns-repeated="16384"/>
</table:table-row>
```

In `src/ods.rs`, these are expanded via loops:

```rust
// read_row() - column expansion
for _ in 0..repeats {
    cells.push(value.clone());
    formulas.push(formula.clone());
}

// get_range() - row expansion
for _ in 0..row_repeats {
    new_cells.extend_from_slice(&row[col_min..]);
}
```

A malicious file can claim billions of repeated cells that overwhelm memory and CPU.

### Expected behavior

No DoS

### Sample code to reproduce

```rust
use calamine::{open_workbook_from_rs, Ods, Reader};
use std::fs::File;
use std::io::BufReader;
use std::time::Instant;

fn main() {
    let file = File::open("malformed-ods-timeout.ods").unwrap();
    let reader = BufReader::new(file);

    let start = Instant::now();
    let result: Result<Ods<_>, _> = open_workbook_from_rs(reader);
    println!("Parsing took: {:?}", start.elapsed()); // ~666ms
}
```

### Test file

[malformed-ods-timeout.ods](https://github.com/user-attachments/files/24418671/malformed-ods-timeout.ods)

### Environment

```text
- calamine version: 0.32.0
- Rust version: stable
- OS: Linux (aarch64)
```

### Checklist

withzombies:
Fix DoS vulnerability where malicious ODS files can exhaust memory by declaring billions of repeated cells via `table:number-rows-repeated` and `table:number-columns-repeated` XML attributes.

 - Add caps matching XLSX limits: 1M rows, 16K columns, 100M total cells
 - Return `OdsError::CellLimitExceeded` instead of silently returning empty range when limits exceeded
 - A 7KB malicious file can no longer attempt to allocate memory for 17+ billion cells

Fixes #594


src/ods.rs
```diff
@@ -27,6 +28,15 @@ use std::marker::PhantomData;
...
 const MIMETYPE: &[u8] = b"application/vnd.oasis.opendocument.spreadsheet";
 
+/// Maximum number of rows allowed in an ODS file (matches XLSX limit).
+const MAX_ROWS: u32 = 1_048_576;
+
+/// Maximum number of columns allowed in an ODS file (matches XLSX limit).
+const MAX_COLUMNS: u32 = 16_384;
+
+/// Maximum number of cells to prevent memory exhaustion from malicious files.
+const MAX_CELLS: usize = 100_000_000;
```
jmcnamara: Minor comment. "MAX_CELLS" might be better as a configurable value in the `Ods` structure but this is fine for now.

tests/test.rs
```diff
@@ -2574,3 +2574,61 @@ fn test_xlsx_table_insertrow_attribute() {
 
     assert!(result.is_ok(), "Expected table_by_name to succeed");
 }
+
```
jmcnamara: I get that you are trying to build out a self describing test case but for the integration tests it would be better to just follow the format of the other tests and add a test against a known problematic file.

This format could possibly be added as a Ods unit test without the zipping.

jmcnamara (commented):
Thanks, good work.

All reviews are comments except for the test which should be changed.

jmcnamara:
@withzombies Merged. Thanks.

PR #600, opened 2026-02-13
Filter chunks shorter then 6 bytes that couse panic

7sedam7:
chunks(6) on the ExternSheet data can yield a final chunk shorter than 6 bytes when the XLS file is malformed or truncated. Accessing xti[2..4] and xti[4..] on a short chunk causes a panic:
   `range end index 4 out of range for slice of length 2`

This adds a `.filter(|xti| xti.len() >= 6)` before mapping, so undersized chunks are skipped instead of panicking.

Reproduces with: corrupt/truncated .xls files where the ExternSheet record's data doesn't align to 6-byte boundaries.

Unfortunately, I don't have xls example, one of my users (where i use calamine) got this panic.


jmcnamara:
@7sedam7 Could you rebase this PR to the latest code.

7sedam7:
Hei, sorry for slowness, I see my issue is now resolved in upstream, so there is nothing to merge, i can close this PR, thanks :)

PR #615, opened 2026-03-02
refactor(cfb): resolve special case in the caller of fn get_chain

sftse:
Follow up on #609 where the confusion is resolved by `len == 0` being caused by a special case in the spec.

Resolving this special case in the caller streamlines the logic inside `fn get_chain`

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -107,7 +107,8 @@ impl Cfb {
 
         // get the list of directory sectors
         debug!("load directories");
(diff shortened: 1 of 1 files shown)


sftse:
Ok, because the code now no longer has a reference to the spec justifying this change, reposting as comment

```text
[MS-CFB] 2.2 Compound File Header
If Major Version is 3, the Number of Directory Sectors MUST be zero.
```

jmcnamara:
Thanks for making the requested changes. Merged.

PR #618, opened 2026-03-05
Book stream

sftse:
Ran across a file with an xls extension that is opened by LibreOffice without warnings, has a single cfb stream called `/BOOK`.

Constructed a synthetic spreadsheet with a renamed stream, can verify opens without errors in LibreOffice.


jmcnamara:
Merged. Thanks.

PR #620, opened 2026-03-06
bench: change to criterion for benchmark testing

jmcnamara:
Change to using the `criterion.rs` crate for `bench` benchmark testing. This avoids the need for `+nightly` to run the benchmarks.

As a side effect it is now possible to run `cargo clippy --all-targets` without `+nightly`.

`criterion.rs` is added as a `dev-dependency`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Clippy checks
       if: ${{ matrix.toolchain == 'stable' }}
       run: |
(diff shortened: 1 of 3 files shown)


jmcnamara:
@jqnatividad for comment.

@ddimaria, note this change since it will have an effect on your PR if merged.

sftse:
I remember divan being a bit lighter on compile time and dependencies than criterion, with a simpler api, at the cost of one feature that I missed, p-value calculation.

jqnatividad:
+1 on criterion.

Issue #625, opened 2026-03-16
Feature required: Replace use of `(u32, u32)` with a type that specifies what is row and column


Erik1000:
Using `(u32, u32)` has no information whether the first is the row index or the second. IMO this would improve UX. Also it if there was a struct, handling of excel native coordinates like `A1` can be supported via converting into this struct.

Currently, I have a lot of helper functions that convert between (u32, u32) and excel coordinates and just messed up with row vs column.

Erik1000:
Also, it would clarify the indexing on `Range`.

jmcnamara:
The `rust_xlsxwriter` crate uses type aliases for rows and columns. This makes the APIs clearer and also helps avoid row/column swapping issues internally. I don't know if that is something like what you are looking for or if it would work for Calamine.

```rust
/// Integer type to represent a zero indexed row number. Excel's limit for rows
/// in a worksheet is 1,048,576.
pub type RowNum = u32;

/// Integer type to represent a zero indexed column number. Excel's limit for
/// columns in a worksheet is 16,384.
pub type ColNum = u16;

// Example usage:
pub fn write_string(&mut self, row: RowNum, col: ColNum, string: impl Into<String>) -> Result<...>
```

However, changing `(u32, u32)` to (effectively) `(u32, u16)` would be an API change.

Erik1000:
Yes, I know it would be a breaking change, but I think it would be worth it.
using u32 and u16 seems better cuz you cant mix them up (tho there are cases where you have to convert from usize).

Since it would be a breaking change anyways, why not use a wrapper struct for RowNum and ColNum together with a Position/Coordinate struct that replaces the tuple?

jmcnamara:
> Since it would be a breaking change anyways, why not use a wrapper struct for RowNum and ColNum together with a Position/Coordinate struct that replaces the tuple?

That would be a much more breaking change. Most end users wouldn't notice a change from `(u32, u32)` to `(u32, u16)` (i.e., their code would compile unchanged) but changing  `(u32, u32)` to a type would break everyones code that uses any API that takes or returns `(u32, u32)`, which is quite a few.

I am sympathetic to having a clear type for `(u32, u32)` but it is probably too late to change that now. Changing to `(RowNum, ColNum)` might be feasible and worth the change for the sake of clarity and maintenance but it would need consensus from other maintainers/committers.

jmcnamara:
Added PR #627 for this.

jmcnamara:
[...]

PR #627, opened 2026-03-20
lib: add RowNum and ColNum types

jmcnamara:
Replaced various row and column `(u32, u32)` instances with type aliases `RowNum` and `ColNum` which are `(u32, u16)`. This makes the API types self documenting and helps avoid row/column mistakes internally.

This is in response to feature request #625. I implemented it mainly to see what would be  required to do this. I don't have a strong desire to add this but on balance it does improve the documentation and also may make the internals a little clearer.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -8,7 +8,7 @@ use serde::{forward_to_deserialize_any, Deserialize, Deserializer};
 use std::marker::PhantomData;
 use std::{fmt, slice, str};
 
-use super::{CellErrorType, CellType, Data, Range, Rows};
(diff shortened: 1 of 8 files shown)


Erik1000:
Perhaps we can also use a type alias `pub type Position = (RowNum, ColNum)` in the code/and docs. this wouldnt introduce more breakage

jmcnamara:
> Perhaps we can also use a type alias `pub type Position = (RowNum, ColNum)` in the code/and docs.

Just my opinion, I think that is slightly less clear and less self documenting.

jmcnamara:
This didn't get any support from other maintainers or users so I will withdraw it.

PR #631, opened 2026-03-25
Allow (and ignore) v nodes in an inlineStr cell.

aquasync:
Test spreadsheet comes from the R readxl package test suite. Returns a parse error without this change.

Cell that seems to cause the problems looks like this:

```
  <c r="D2" s="2" t="inlineStr">
    <v>1.0</v>
    <is>
      <r>
        ...
        <t xml:space="preserve">1.</t>
      </r>
    </is>
  </c>
```

It appears in excel with a number-formatted-as-text tooltip.


aquasync:
Not 100% sure that just ignoring it is the right approach - happy to hear other opinions. Also I'm not familiar enough with the standard to know if the last value that appears is the one that should take precedence? If the `<v>1.0</v>` node above came after the `<is>` one then I think this fix would make the cell blank, as the outer loop seems to just take the last value that it sees.

jmcnamara:
PRs that fix a bug should start with a bug report. That allows for an evaluation of the error and (if necessary) a discussion about what the fix should be. It also allows the PR to refer to the bug it is fixing. 

Could you do that now. Also for your other report.

aquasync:
Sure will do. Failing format checks so will fix that too.

jmcnamara:
I think this would be better handled at a level above `read_v()` in `read_value()`. Like this:

```diff
diff --git a/src/xlsx/cells_reader.rs b/src/xlsx/cells_reader.rs
index 699f9a4..5925129 100644
--- a/src/xlsx/cells_reader.rs
+++ b/src/xlsx/cells_reader.rs
@@ -346,6 +346,13 @@ where
             read_string_with_bufs(xml, e.name(), &mut bufs.xml, &mut bufs.str_inner)?
                 .map_or(DataRef::Empty, DataRef::String)
         }
+        // Ignore <v> for inlineStr cells since it is redundant. The value is in
+        // the <is> element, which is handled above.
+        b"v" if matches!(type_attr, Some(b"inlineStr") | Some(b"is")) => {
+            bufs.xml.clear();
+            xml.read_to_end_into(e.name(), &mut bufs.xml)?;
+            DataRef::Empty
+        }
         b"v" => match type_attr {
             Some(b"n") | Some(b"s") | Some(b"b") | Some(b"e") | None => {
                 // These types are always plain ASCII (no CR/LF or entities), so we can
@@ -442,13 +449,6 @@ fn read_v<'s>(
                     }
                 })
         }
-        Some(b"is") => {
-            // this case should be handled in outer loop over cell elements, in which
-            // case read_inline_str is called instead. Case included here for completeness.
-            Err(XlsxError::Unexpected(
-                "called read_value on a cell of type inlineStr",
-            ))
-        }
         Some(t) => {
             let t = std::str::from_utf8(t).unwrap_or("<utf8 error>").to_string();
             Err(XlsxError::CellTAttribute(t))
```

This also improves/merges the inner `<is>` handling where it says "this case should be handled in outer loop over cell elements".

tests/test.rs
```diff
@@ -2869,3 +2869,15 @@ fn test_capitalized_book_stream() {
     // ref: https://github.com/tafia/calamine/issues/618
     let _wb: Xls<_> = wb("capitalized_wbook_stream.xls");
 }
+
+#[test]
```
jmcnamara, line 2873: The test should have a comment to say what it is testing and include the # number for the bug report. Also give the test case and test file more descriptive names like `test_xlsx_inline_str_with_value()` and `inlineStr_with_value.xlsx`

jmcnamara (changes requested):

jmcnamara:
The commit message would be better as something like this:

```
xlsx: fix issue with redundant <v> value in inlineStr

Closes #633
```

You can decide if it merits a longer explanation in the body of the message. Have a look at the better examples in the git log.

aquasync:
Sure thing, have updated.

jmcnamara:
Thanks. That looks good.

One other note on coding standards for future patches: Comments should have proper sentence formatting and punctuation. 

However, this is fine, so I will merge it as is.

PR #639, opened 2026-03-27
xlsx: handle minimal package structure

aquasync:
The code currently assumes the workbook xml (and related files) live in `xl/`, however this needn't be the case. It further assumes the directory names can be relied upon to infer sheet types.

The first is replaced with reading the document path from `_rels/.rels`, and the latter from the types in the `_rels/workbook.xml.rels`.

The code for the rels xml parsing seems a bit clunky; maybe cleaner way to do it with quick_xml, I was just following the approach used elsewhere.

I also considered just ignoring and trying extra known paths (eg "xl/workbook.xml", then "workbook.xml"), however it won't work with something like: [arbitrarypath.xlsx](https://github.com/user-attachments/files/26306248/arbitrarypath.xlsx).


src/xlsx/mod.rs
```diff
@@ -1653,6 +1704,7 @@ impl<RS: Read + Seek> Reader<RS> for Xlsx<RS> {
     }
 
     fn worksheet_range(&mut self, name: &str) -> Result<Range<Data>, XlsxError> {
+        dbg!("here!");
```
jmcnamara: Remove this debug statement.

src/xlsx/mod.rs
```diff
@@ -438,28 +504,31 @@ impl<RS: Read + Seek> Xlsx<RS> {
...
+                                rel_id = (*v).to_vec();
                             }
                             _ => (),
                         }
                     }
-                    let typ = match path.split('/').nth(1) {
-                        Some("worksheets") => SheetType::WorkSheet,
-                        Some("chartsheets") => SheetType::ChartSheet,
-                        Some("dialogsheets") => SheetType::DialogSheet,
+                    let (r, rel_type): &&(_, String) = &relationships
```
jmcnamara: I think this could be written more simply as 

```rust
let (r, rel_type) = relationships
      .get(&rel_id)
      .ok_or(XlsxError::RelationshipNotFound)?;
```

src/xlsx/mod.rs
```diff
@@ -410,7 +476,7 @@ impl<RS: Read + Seek> Xlsx<RS> {
             match xml.read_event_into(&mut buf) {
                 Ok(Event::Start(e)) if e.local_name().as_ref() == b"sheet" => {
                     let mut name = String::new();
-                    let mut path = String::new();
+                    let mut rel_id: Vec<u8> = b"".to_vec();
```
jmcnamara: Just use `Vec::new()` here: `let mut rel_id = Vec::new()`.

src/xlsx/mod.rs
```diff
@@ -247,6 +247,8 @@ type Tables = Option<Vec<(String, String, Vec<String>, Dimensions)>>;
 /// Xlsx, Xlsm, Xlam
 pub struct Xlsx<RS> {
     zip: ZipArchive<RS>,
+    /// Path to zip directory where excel xml is
```
jmcnamara: Comments should be complete sentences with proper punctuation. Something like `Path to xml file within zip archive.` would be better here. It **may** also be worth explaining that the path always ends in / or is empty since all of the format statements are like this: `format!("{}foo.xml", self.xl_path)`.

tests/test.rs
```diff
@@ -2920,74 +2883,13 @@ fn test_xlsx_inline_str_with_value() {
     assert_eq!(range.get_value((1, 3)), Some(&String("1.".to_string())));
 }
 
-// Test for issue #648 where a inlineStr string contains CDATA.
 #[test]
```
jmcnamara, line 2995: You will need to rebase/resolve this test function since it is in conflict with the current main.

jmcnamara (changes requested):

jmcnamara:
Some other comments.

There are two or more places where the path replacements were missed:

The  `read_pictures()` function which is behind a feature flag:

`if zname.starts_with("xl/media") {`

If the workbook lives outside xl/, pictures won't be found. Should be something like:

`if zname.starts_with(&format!("{}media", self.xl_path)) {`

And similarly in the pivot table code:

```rust
  if target.starts_with("xl/pivotCache") {
      record_path.replace(target);
  } else if !target.is_empty() {
      record_path.replace(format!("xl/pivotCache/{target}"));
  }
```

Do a grep for other `"xl/"` places in the code to ensure that they are handled correctly.

Also, is making `_rels/.rels` mandatory a breaking change? The  `read_package_relationships()` returns `Err(XlsxError::FileNotFound)` if the file is absent. The old code (rightly or wrongly) silently assumed `xl/`. Is this in accordance with the spec or are there any known xlsx files in the wild (in that R test suite you are working through) that lack `_rels/.rels`? If so, this is a regression, which may be okay, but we should be aware of it so it can be documented in the release notes.

aquasync:
Thanks for going through this. I've addressed your suggestions and added a new commit which tackles the media & pivot cache.

As for other uses of `"xl/"`, from what I can see the only remaining place is actually in `read_workbook`, where it potentially strips it if leading. I'd previously left that alone as the purpose seemed unclear, but I think it is handling the fact that paths can be either relative or absolute. As such stripping an initial `"/xl/"` makes sense, but not `"xl/"` (that should resolve to a path like `"xl/xl/..."`).

I changed the behaviour to what I think it should be doing there - leaving the path alone if it starts with `"/"`, otherwise treating as relative to workbook xml (ie `xl_path`). More broadly I think some sort of `path_combine` helper is probably needed in other places to handle resolving relative vs abs paths, though that'd be a bigger change. 

I created a more pathological file here to test: [test-notxl-elsewhere.xlsx](https://github.com/user-attachments/files/27699178/test-notxl-elsewhere.xlsx). It opens and displays correctly in Excel & Libreoffice. It has the workbook xml in a directory `/notxl/`, and the sheets xml in a different directory `/elsewhere/`, referred to by absolute path. It can now be parsed correctly with this set of changes.

As to your other point, that's a valid concern. There's no file in the R test suite missing `_rels/.rels`, and the reader code requires its existence. As a test I tried deleting it from a test xlsx, and both Excel & Libreoffice complain the file is corrupt and offer to repair (neither successfully). Google sheets does display the data though. Pretty sure it is non-conforming, but I guess in the spirit of Postel we could easily restore current behaviour? Just a matter of returning `Ok(())` instead in a few places, as the default `xl_path` is already set.

aquasync:
Hmm, ran into one other issue with a file that had a leading slash: `<Relationship ... Target="/xl/workbook.xml" ...>`. That caused `xl_path` to be `"/xl/"` and broke subsequent reads.

jmcnamara:
> As to your other point, that's a valid concern. There's no file in the R test suite missing `_rels/.rels`, and the reader code requires its existence.

I also tried several tests/tools with a file without `_rels/.rels` and it looks like it isn't valid. So ignore that constraint.

jmcnamara:
Merged. Thanks.

PR #643, opened 2026-04-24
xls: fix panic parsing formulas in BIFF5 workbooks

mogery:
parse_formula decoded the non-3D PtgRef (0x24/0x44/0x64), PtgArea (0x25/0x45/0x65), PtgRefErr (0x2A/0x4A/0x6A), and PtgAreaErr (0x2B/0x4B/0x6B) tokens assuming BIFF8 sizes (4 and 8 data bytes respectively). In BIFF2-5 those tokens use 1-byte columns, so their data blocks are 3 and 6 bytes. When a BIFF5 workbook hit one of these tokens, the parser advanced rgce by too many bytes, mis-identified following bytes as a new Ptg, and eventually panicked with
"range end index 2 out of range for slice of length 0" inside utils::read_u16 - taking down any caller that opened the file.

Thread biff through parse_formula and branch the four handlers on the workbook's BIFF version, mirroring how parse_defined_names and the 3D Ptg handlers already do it.

Example file that caused a panic without this patch: https://www.asx.com.au/content/dam/asx/participants/derivatives-market/equity-derivatives/equity-derivatives-statistics/2016/annual-market-summary-2016.xls

PR authored with Claude Opus 4.7


mogery:
Apologies, fixed the formatting.

jmcnamara:
Thanks. Could you also add a test case to the `tests/test.rs` file with the sample xls file that demonstrates the issue.

mogery:
Added a test with a minimal fixture file. Panics on master, works on this branch, opens in Excel cleanly.

PR #645, opened 2026-04-24
xlsx: fix shared formula expansion corrupting function names

### LINKED ISSUE #644: Bug: shared formula expansion corrupts function names that parse as valid cell references (e.g., LOG10 -> LOG11)
### Current behavior

When expanding shared formulas, `replace_cell_names()` in `src/xlsx/mod.rs` treats `LOG10` as a cell reference (column `LOG` = #8509, row `10`) and applies the row offset. This produces `LOG11`, `LOG12`, ..., `LOG19` for subsequent rows in the shared formula group.

```
Shared formula in B1: LOG10(ABS(A1)+1)

calamine returns:
  B1:  LOG10(ABS(A1)+1)   <- correct (master formula)
  B2:  LOG11(ABS(A2)+1)   <- wrong: LOG10 -> LOG11
  B3:  LOG12(ABS(A3)+1)   <- wrong: LOG10 -> LOG12
  ...
  B10: LOG19(ABS(A10)+1)  <- wrong: LOG10 -> LOG19
```

`LOG11` through `LOG19` are not valid Excel functions. The formula text is corrupted.

### Expected behavior

Function names should not be modified by shared formula expansion. Only cell references should be offset.

```
  B1:  LOG10(ABS(A1)+1)
  B2:  LOG10(ABS(A2)+1)   <- only A1 -> A2 adjusted
  B3:  LOG10(ABS(A3)+1)   <- only A1 -> A3 adjusted
```

### Sample code to reproduce

Using the attached `log10_shared.xlsx`:

```rust
use calamine::{open_workbook, Reader, Xlsx};

fn main() {
    let mut workbook: Xlsx<_> = open_workbook("log10_shared.xlsx").unwrap();
    let formulas = workbook.worksheet_formula("Sheet1").unwrap();

    for (row_idx, row) in formulas.rows().enumerate() {
        for (col_idx, cell) in row.iter().enumerate() {
            let text = cell.to_string();
            if text.contains("LOG") {
                println!("row={row_idx} col={col_idx}: {text}");
            }
        }
    }
    // Expected: all rows show LOG10(...)
    // Actual: row 0 shows LOG10, rows 1+ show LOG11, LOG12, ...
}
```

### Test file

Attached above: `log10_shared.xlsx` — 10 rows with `=LOG10(ABS(A{row})+1)` as a shared formula.

### Root cause

`Reference::parse()` at `src/xlsx/mod.rs:2046` accepts `LOG10` as `Reference::Cell { col: 8508, row: 9 }` because:
- `LOG` is a valid column name (column #8509, within Excel's 16,384 limit)
- `10` is a valid row number

`replace_cell_names()` at line 2246 passes each alphanumeric token to `offset_range()` -> `Reference::parse()`. Since `LOG10` parses successfully, the row gets offset.

`LOG10` is the only commonly affected Excel function name — `ATAN2`'s column letters (`ATAN` = #31,136) exceed the 16,384 column limit so `validate()` rejects it. Other digit-containing function names (`HEX2DEC`, `BIN2DEC`, etc.) have non-alpha characters that break the alphanumeric token before the digits.

### Suggested fix

In `replace_cell_names()`, skip tokens that are immediately followed by `(` — those are function calls, not cell references:

```rust
// line 2258 (mid-loop token flush)
if token_start < token_end {
    let next_is_paren = c == b'(';
    if next_is_paren
        || offset_range(&bytes[token_start..token_end], offset, &mut res).is_err()
    {
        res.extend(&bytes[token_start..token_end]);
    }
}
```

The post-loop flush (line 2273) needs no change — a token at end-of-string cannot be a function call.

### Environment

- calamine version: 0.34.0
- Cargo.toml dependency line: `calamine = "0.34"`
- rustc version: 1.85.0
- OS: macOS 15.5

cilladev:
Fixes #644.

## What

`replace_cell_names()` treats `LOG10` as a cell reference (column `LOG` = #8509, row `10`) and applies the row offset during shared formula expansion. This produces `LOG11`, `LOG12`, ..., `LOG19` for subsequent rows.

## Why

`Reference::parse()` accepts `LOG10` because `LOG` is a valid Excel column name (within the 16,384 column limit) and `10` is a valid row number. The tokenizer in `replace_cell_names()` cannot distinguish function names from cell references.

## How

Skip tokens immediately followed by `(` — in Excel formula syntax, a `(` after an alphanumeric token always means function call, never cell reference.

## Testing

```
$ cargo run --example try
# Before: row 1 shows LOG11, row 2 shows LOG12, ...
# After:  all rows show LOG10
```

Test file (`log10_shared.xlsx`) attached to #644.


jmcnamara:
Thanks. Could you also add a test case to the `tests/test.rs` file with the sample xls file that demonstrates the issue.

PR #646, opened 2026-04-25
xls: fix unsupported formula function panic

mogery:
Handle unsupported XLS function ids as formula parse errors instead of indexing past the function table, and cover issue with a valid minimal XLS fixture.

Example real file that causes a panic: https://www.wmfs.net/wp-content/uploads/2022/03/Waste-Contracts.xls

Patch written with GPT-5.5


mogery:
couple of force pushes to fix excel complaining about the fixture file

jmcnamara:
Merged manually to main to resolve conflict.

Issue #486, opened 2025-01-28
problems getting hyperlink info and images from specified worksheets


nathanaelhuffman:
First off, thanks for this library I've been using it in a number of places to deal with xlsx files of varying types, being able to loop rows and pass stuff to serde etc has all been really awesome!

I did run into a stumbling block on a recent project and had to bail out to another library to deal with some of the features I needed as I couldn't get this information out just using calamine.

I unfortunately have an xlsx template that has a varying number of images pasted onto different sheets, and hyperlinks on the first sheet that link
to the corresponding image. Each of these images is associated with some text in cells on that first sheet, think issue tracker with issue text and metadata on the first sheet linking to supporting images on subsequent sheets.

Problem 1: I can't read the raw hyperlink info in the link cells on the original sheet.  The hyperlinks in the  cell have a display text set to "Image" but the link links to <SHEET>!<CELL> so like "Sheet 2!A1", but calamine only shows Some("Image") which is the text representation for these cells.  This is problematic as I can't figure out where the link goes.  Maybe related to #360 ?

Problem 2: the image support in calamine it far too simplistic for anything but basic image export. Even with the "pictures" feature, while you can get the image data, the only metadata you get with it is its image type (jpg, png etc), no information about the image name in the zip, no information about which cells reference this etc.

Both of these problems conspire to make it so that I'm unable to figure out which image is on which sheet, and which issue links to which sheet (and thus which image) with just calamine alone.

For a possible implementation reference, I really like calamine's API for most of my other processing but I bailed out to [umya-spreadsheet](https://github.com/MathNya/umya-spreadsheet)
and used the get_hyperlink function on their [cell type](https://github.com/MathNya/umya-spreadsheet/blob/e5f44aeeda8ae3e4e8a1f7fad83311bfe0e5a801/src/structs/cell.rs#L105) to get the hyperlink info on the link cells and then used [get_image](https://docs.rs/umya-spreadsheet/latest/umya_spreadsheet/structs/struct.Image.html) to get the image data from the specified link cells.  

It would be really awesome to have similar API functionality in calamine!

jmcnamara:
I modified the title so that it only refers to one feature request to make tracking implementation easier.

Hyperlinks are covered by feature request: #360

jmcnamara:
Closed (for picture information) in favour of #652 and for hyperlinks #360.

Issue #381, opened 2023-12-04
Pictures information


schungx:
Is it possible to get information on embedded pictures, such as location (x, y),  which cell is it inside (or more than one cell), width x height, scaling and so on?

Currently it seems the only information available is its extension.

nathanaelhuffman:
This certain feels like something that would be useful to me, especially knowing which cell/sheet the image is on.

jmcnamara:
Closing in favour of #652.

PR #652, opened 2026-05-12
picture: extend the picture feature to include more image metadata

jmcnamara:
Extend the pictures feature to extract cell position information, the raw image data and to also to handle newer Excel 365 style embedded images.

This incorporates PR #592 and the suggested changes. The original author is credited in the commit.

Relates to: 
- #592
- #486 
- #381


jmcnamara:
TODO:

- [x] Needs example in `examples` directory.
- [x] Needs test case with mixed legacy and embedded images.

PR #628, opened 2026-03-23
Implement conditional formatting

ddimaria:
## Add conditional formatting support for XLSX

Adds read support for conditional formatting rules from XLSX worksheets. This parses `<conditionalFormatting>` elements and their child `<cfRule>` nodes, resolving differential format styles (`dxf`) into the existing `Style` type.

**Branched from https://github.com/tafia/calamine/pull/653, so this should be reviewed after 653 is merged into main.**

### Supported rule types

| Rule type | Variant |
|---|---|
| Cell value comparison | `CellIs` (all operators: equal, between, greaterThan, etc.) |
| Color scales | `ColorScale2`, `ColorScale3` |
| Data bars | `DataBar` (with fill and border colors) |
| Icon sets | `IconSet` (all 20 built-in icon set types) |
| Top/Bottom N | `Top10` (rank, percent, bottom) |
| Above/below average | `AboveAverage` (with std deviation support) |
| Text rules | `Text` (contains, notContains, beginsWith, endsWith) |
| Time period | `TimePeriod` (yesterday, today, last7Days, etc.) |
| Formula-based | `Expression` |
| Blanks/errors | `ContainsBlanks`, `NotContainsBlanks`, `ContainsErrors`, `NotContainsErrors` |
| Duplicates/uniques | `DuplicateValues`, `UniqueValues` |
| Fallback | `Unknown` (preserves raw type string) |

### API

Two new public methods on `Xlsx<RS>`:

- `worksheet_conditional_formatting(&mut self, name: &str)` -- by sheet name
- `worksheet_conditional_formatting_at(&mut self, index: usize)` -- by sheet index

Both return `Vec<ConditionalFormatting>`, where each block contains the target `sqref` range and a list of prioritized rules with resolved `Style` formats.

### Changes

- **`src/conditional_format.rs`** (new): Public data model -- `ConditionalFormatting`, `ConditionalFormatRule`, `ConditionalFormatRuleType`, and supporting enums (`CfOperator`, `CfTextOperator`, `CfValueObject`, `IconSetType`, `TimePeriodType`, etc.)
- **`src/xlsx/mod.rs`**: Parses `<dxfs>` from `styles.xml` into differential format styles; adds `parse_conditional_formattings` and helpers (`parse_cf_rules`, `parse_single_cf_rule`, `parse_color_scale`, `parse_data_bar`, `parse_icon_set`, `parse_cfvo`) to walk the worksheet XML
- **`src/lib.rs`**: Re-exports all new public types
[...]

PR #653, opened 2026-05-12
feat(xlsx): implement cell style extraction with rich text and worksheet layout

ddimaria:
## Summary

Adds style support for Calamine xlsx files. Supersedes #538, which had stale review comments and a long discussion thread that made the page hard to load. This PR is a clean, single squashed commit on top of `master`.

## Public API

```rust
// row × col grid of cell styles (RLE-compressed internally)
let styles = xlsx.worksheet_style("Sheet1")?;
for (row, col, style) in styles.cells() {
    if let Some(font) = style.get_font() {
        if font.is_bold() { /* ... */ }
    }
}

// column widths and row heights
let layout = xlsx.worksheet_layout("Sheet1")?;
```

## What's included

- `src/style.rs`: `Style`, `Color`, `Font`, `Fill`, `Borders`, `Alignment`, `NumberFormat`, `Protection`, `RichText`, `TextRun`, `StyleRange`, `WorksheetLayout`.
- `src/xlsx/style_parser.rs`: parser for `styles.xml`, theme colors with tint resolution, indexed-color fallback, `sysClr` last-resolved fallback.
- Run-length encoded `StyleRange` for memory efficiency on large workbooks (tested with a 1M-cell fixture).
- Rich text support: the shared-string reader now decodes `<r>` runs and preserves per-run formatting. Also handles the case (consistent with #637) where rich runs follow an initial plain `<t>`.
- Benchmarks in `benches/style.rs` (criterion).
- Test fixtures: `styles.xlsx`, `borders.xlsx`, `EMSI_JobChange_UK.xlsx`, `problematic_formats.xlsx`, `styles_1M.xlsx`.

## Notes on review feedback from #538

- Squashed into a single commit per @jmcnamara's request.
- Removed `STYLE_FEATURE.md` (was flagged for removal in the previous review).
- Removed accidentally committed profiler artifact (`profile.json.gz`).
- `xlsb` support and conditional formatting are intentionally left to follow-up PRs (#628 already exists for conditional formatting).

## Test plan

- [x] `cargo test --all-features` — 267 tests pass (159 integration, 63 doc, 45 unit), 0 failures.
- [x] `cargo check --tests --all-features` — clean (warnings are pre-existing).
- [ ] `cargo fmt --check` / `cargo clippy` — happy to address any remaining lints during review.


ddimaria:
@jmcnamara I just pushed up a lint fix for this one

benches/generate_large_styled_xlsx.rs
```diff
@@ -0,0 +1,206 @@
+// SPDX-License-Identifier: MIT
```
jmcnamara: Rather than generate this file dynamically, and require another dependency (even if it is a dev one) it would be simpler to just generate the required files and add them to the `tests` directory. 

Also, this doesn't matter for a benchmark test case, but I think Excel only supports 32k styles per workbook.

benches/generate_styles_1M.rs
```diff
@@ -0,0 +1,206 @@
+// SPDX-License-Identifier: MIT
```
jmcnamara: Same comment as other benchmark test case, generate the file and add it to the tests directory.

examples/read_row_and_column_dimensions.rs
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright 2016-2025, Johann Tuffe.
+
+use calamine::{open_workbook, Reader, Xlsx};
+
+/// Example demonstrating how to capture column widths and row heights from Excel files
```
jmcnamara, line 7: Give this example a better name like `read_row_and_column_dimensions.rs`. Also add it to the `examples/README.md` file.

examples/layout.rs
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright 2016-2025, Johann Tuffe.
+
+use calamine::{open_workbook, Reader, Xlsx};
+
+/// Example demonstrating how to capture column widths and row heights from Excel files
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Open an Excel file
+    let path = format!("{}/tests/styles.xlsx", env!("CARGO_MANIFEST_DIR"));
```
jmcnamara: Don't bother to use `env!("CARGO_MANIFEST_DIR")` in the examples since it isn't necessary. Just use the explicit path instead: `let path = "tests/styles.xlsx";`

jmcnamara (changes requested):

jmcnamara:
@ddimaria I will try to do a high level review of this over the next few days/weekend.

You will need to rebase/resolve conflicts due to the commits that will be merged ahead of this.

benches/generate_large_styled_xlsx.rs
ddimaria: Good call. Removed this generator and committed the generated file directly at tests/styles_1M.xlsx, so there's no runtime generation step. I've also dropped the rust_xlsxwriter dev-dependency entirely since it was only used here. And thanks for the note on the ~32k styles-per-workbook limit.

benches/generate_styles_1M.rs
ddimaria: Done. This one was actually a duplicate of the other generator (both wrote tests/styles_1M.xlsx). The style benchmark now reads the committed tests/styles_1M.xlsx directly.

examples/read_row_and_column_dimensions.rs
ddimaria: Renamed to examples/read_row_and_column_dimensions.rs and added it to examples/README.md.

examples/layout.rs
ddimaria: Done — the example now uses let path = "tests/styles.xlsx"; directly.

ddimaria:
> @ddimaria I will try to do a high level review of this over the next few days/weekend.
> 
> You will need to rebase/resolve conflicts due to the commits that will be merged ahead of this.

Conflicts resolved @jmcnamara

jmcnamara:
@ddimaria 

Thanks for resolving the conflicts. I did the same thing in #682 in order to review the Styles changes without (for now) the rich text and Layout code. Could you have a look at that and compare against your version. I made the changes I would have requested from yours.

ddimaria:
@jmcnamara Following up on this: I've rebased `feat/styles` onto the current `master` (post quick-xml 0.42 and the new raw-attribute handling), so the PR now applies cleanly and is conflict-free again. The earlier inline comments are all addressed as noted above: the benchmark generators were removed in favor of a committed `tests/styles_1M.xlsx`, the `rust_xlsxwriter` dev-dependency was dropped, the example was renamed to `read_row_and_column_dimensions.rs` and added to `examples/README.md`, and the examples now use plain relative paths instead of `env!("CARGO_MANIFEST_DIR")`.

I also had a look at #682 — nice cut-down, and I like `next_cell_with_style_id()` for the single-pass use case. The main things this PR carries beyond it are the rich-text shared-string support and the worksheet layout (column widths / row heights) API. Happy to go whichever way is easier for you: continue reviewing this PR as-is, or I can rebase the rich text and layout pieces on top of #682 once it lands.

jmcnamara:
@ddimaria Thanks for your patience. 

> Happy to go whichever way is easier for you: continue reviewing this PR as-is, or I can rebase the rich text and layout pieces on top of https://github.com/tafia/calamine/pull/682 once it lands.

At the moment I'm leaning towards integrating #682 and then taking your Rich Text and Column/Row work on top of that as separate commits/PRs. And then after that your other PRs.

I hope to get back to this in the next week or two.

Issue #247, opened 2021-12-31
Associating data and formula


KillTheMule:
There are iterators over cell data as well as the formulas behind the data, but there does not seem to be a way to associate those. By that I mean that

```rust
workbook.worksheet_formula(&sheetname).rows().next().unwrap().get(0)
```
is not necessarily the formula that produced the data

```rust
workbook.worksheet(&sheetname).rows().next().unwrap().get(0)
```

but rather something like "the formula of the first cell in the first row that contains a formula at all" or something similar. The "culprit" seems to be the [independent call of `get_range`](https://github.com/tafia/calamine/blob/master/src/ods.rs#L259) for values/formulas.

Is there a way to recover that information? Would you be open to change that code? It would of course be some sort of breaking change of the worst kind (stuff still compiles, but behaves differently), but it seems to me that it would be pretty useful for others, too, to be able to associate data and formulas.

jmcnamara:
Would be closed by #654

jmcnamara:
Fixed in #654.

KillTheMule:
Nice! Beat me though, but isn't this just for xlsx? My personal use case would have involved ods, but it's not actually a pressing matter anymore.

PR #654, opened 2026-05-12
feat(xlsx): add XLSX cell+formula streaming records with shared formula metadata

PSU3D0:
## Summary

Adds XLSX-specific streaming APIs for reading cell values and formulas in a single worksheet XML pass:

- `XlsxCellReader::next_cell_with_formula()`
  - yields value + expanded per-cell formula text
  - preserves convenience semantics for consumers that want formula text directly

- `XlsxCellReader::next_cell_with_formula_metadata()`
  - yields value + semantic formula metadata
  - exposes shared formula anchors and derived cells without expanding every derived formula
  - avoids per-derived-cell formula translation/allocation for span-aware consumers

Also adds shared-formula translation helpers:

- `translate_formula(...)`
- `translate_formula_into(...)`

`XlsxFormulaMetadata` is marked `#[non_exhaustive]` for future XLSX formula kinds.

## Motivation

Current consumers that need both values and formulas must either use separate APIs or expand shared formulas eagerly. The metadata path lets consumers preserve XLSX shared-formula structure and decide whether/when to expand.

In particular, this work came from improving the Calamine-backed XLSX load path in [formualizer](https://github.com/PSU3D0/formualizer), which improved load times in our benchmarking by ~125ms absolute on various 100k row workloads.

This is useful for consumers that can ingest shared formula families/spans directly.

## Compatibility

This does not change existing public APIs. The new APIs are additive and XLSX-specific.

## Benchmarks

Synthetic XLSX with shared formulas:

```text
xlsx_shared_formula_cells_reader/expanded/100000  ~152.16 ms
xlsx_shared_formula_cells_reader/metadata/100000  ~133.30 ms
```

Allocator-counting harness on 100k shared formulas:

```text
expanded:
  alloc_calls: 1,300,111
  alloc_bytes: 105,800,269

metadata:
  alloc_calls:   800,121
  alloc_bytes: 100,122,568
```

RSS is effectively unchanged because XLSX/XML buffers dominate peak memory, but the metadata path significantly reduces allocation churn.

Validation

```bash
cargo fmt --all -- --check
cargo test --all-targets
cargo clippy --all-targets -- -D warnings
cargo bench --bench basic xlsx_shared_formula_cells_reader -- --warm-up-time 1 --measurement-time 3 --sample-size 10
```


jmcnamara:
This looks like a good PR. Thanks. I'll review in full on the next few days. 

Potentially closes #247.

jmcnamara:
> Validation
> 
> ```shell
> cargo fmt --all -- --check
> cargo test --all-targets
> cargo clippy --all-targets -- -D warnings
> ```

That is good. It is rare (unfortunately) that a PR passes the CI first time.

One quick request before I get to a review, could you replace the programmatically generated xlsx files in the test cases with real `*.xlsx` files that should go in the `tests` directory with suitable names. The minimal files are okay if they load without error in Excel but It is probably best to load them in Excel and resave them. It may also be worth changing the unit test to an integration test in `tests/test.rs`.

PSU3D0:
Unfortunately I am running Linux and don't have easy access to a windows machine or Excel license. I understand the value of having the excel re-save so shared-string testing is higher conviction.

Is a LibreOffice re-save a suitable alternative?

jmcnamara:
> Is a LibreOffice re-save a suitable alternative?

For something like this it would be perfectly fine. To save you some time I extracted the test code that creates the test files and generated the xlsx files and resaved them with Excel:

[shared_formula_simple.xlsx](https://github.com/user-attachments/files/27736687/shared_formula_simple.xlsx)
[shared_formula_bench.xlsx](https://github.com/user-attachments/files/27736688/shared_formula_bench.xlsx)

Double check the shared formula structure.

These should go in the `tests` directory.

PSU3D0:
Thanks for generating and Excel-resaving those fixtures - I pushed an update using them.

Changes in the latest commit:
- Added `tests/shared_formula_simple.xlsx` and `tests/shared_formula_bench.xlsx`.
- Moved the shared-formula streaming/metadata coverage into `tests/test.rs` so it exercises real XLSX fixtures through the public reader path.
- Removed the programmatic minimal-XLSX helper/tests for this coverage.

I also double-checked the structure before wiring the tests:
- `shared_formula_simple.xlsx`: `B1` is the `si=0` anchor with `ref="B1:B3"` and `A1*2`; `B2:B3` are shared followers with no formula text.
- `shared_formula_bench.xlsx`: same pattern over `B1:B100`.

Local validation:

```shell
cargo fmt --all -- --check
cargo test --all-targets
cargo clippy --all-targets -- -D warnings
```

PSU3D0:
Just checking in on this PR. Cutting a 0.6.0 release for formualizer this week and would love to include the calamine perf gains in the blog post :slightly_smiling_face: 

No big rush though and I'm not familiar with Calamine's release schedule.

benches/basic.rs
```diff
@@ -7,10 +7,69 @@
...
-use criterion::{criterion_group, criterion_main, Criterion};
+use calamine::{open_workbook, Ods, Reader, Xls, Xlsb, Xlsx, XlsxFormulaMetadata};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use std::fs::File;
-use std::io::BufReader;
+use std::hint::black_box;
+use std::io::{BufReader, Cursor, Write};
+use zip::write::SimpleFileOptions;
+
+fn shared_formula_xlsx(rows: u32) -> Vec<u8> {
```
jmcnamara: This inline file creation should be replace with a reference to the actual shared_formula_bench.xlsx file in the `tests` directory.

src/xlsx/mod.rs
```diff
@@ -2258,10 +2260,60 @@ fn offset_range(range: &[u8], offset: (i64, i64), buf: &mut Vec<u8>) -> Result<(
...
+///
+/// `anchor` and `target` are zero-based `(row, column)` positions. Relative
+/// references in `formula` are shifted by `target - anchor`; absolute
+/// references remain fixed. The translation is the same one used by Calamine
+/// when expanding XLSX shared formulas for [`Reader::worksheet_formula`].
+///
+/// This helper is intentionally workbook-state agnostic. XLSX shared-formula
+/// indices are worksheet-local, so callers are responsible for pairing the
+/// right template/anchor with each derived cell.
+pub fn translate_formula(
```
jmcnamara: The verb "translate" is ambiguous and slightly confusing here. I'd suggest something like `expand_share_formula()`. I'd also suggest `start` and `end` instead of `anchor` and `target`.

src/xlsx/mod.rs
```diff
@@ -2258,10 +2260,60 @@ fn offset_range(range: &[u8], offset: (i64, i64), buf: &mut Vec<u8>) -> Result<(
...
+}
+
+#[cfg(test)]
+fn translate_formula_with_offset(formula: &str, offset: (i64, i64)) -> Result<String, XlsxError> {
+    let mut out = String::with_capacity(formula.len());
+    translate_formula_with_offset_into(formula, offset, &mut out)?;
+    Ok(out)
+}
+
 // Advance all valid cell names in the string by the offset.
```
jmcnamara: Update/improve the comment for the function.

src/xlsx/mod.rs
```diff
@@ -2273,9 +2325,12 @@ fn replace_cell_names(s: &str, offset: (i64, i64)) -> Result<String, XlsxError>
         } else {
             if token_start < token_end {
                 let next_is_paren = c == b'(';
-                if next_is_paren
-                    || offset_range(&bytes[token_start..token_end], offset, &mut res).is_err()
+                token_buf.clear();
+                if !next_is_paren
+                    && offset_range(&bytes[token_start..token_end], offset, &mut token_buf).is_ok()
```
jmcnamara, line 2332: This is good. As far as I can tell this fixes an existing bug in passing (or intentionally). If that is the case them I could be a separate fix commit before the new functionality. However, it is probably okay to have it as part of the the overall patch since the bug wasn't exposed previously.

tests/test.rs
```diff
@@ -1942,6 +1942,99 @@ fn issue_391_shared_formula() {
     assert!(expect.cells().eq(res.cells()));
 }
 
+#[test]
+fn xlsx_cell_reader_expands_excel_resaved_shared_formulas() {
```
jmcnamara, line 1948: Put a 1-2 line comment here to explain what is being tested and refer to this PR #654.

src/xlsx/cells_reader.rs
```diff
@@ -23,7 +22,11 @@ use crate::{
     Cell, XlsxError,
 };
 
-type FormulaMap = HashMap<(u32, u32), (i64, i64)>;
+#[derive(Clone, Debug)]
+struct SharedFormulaTemplate {
```
jmcnamara: Call this `SharedFormula` and rename `anchor` to `range`

src/xlsx/cells_reader.rs
```diff
@@ -49,6 +52,117 @@ impl ValueBufs {
...
+#[non_exhaustive]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum XlsxFormulaMetadata {
+    /// Ordinary, non-shared formula.
+    Normal {
+        /// Formula text.
+        formula: String,
+    },
+    /// Shared formula anchor/template cell.
+    SharedAnchor {
```
jmcnamara: Call this variant `Shared` instead of `SharedAnchor`.

src/xlsx/cells_reader.rs
```diff
@@ -49,6 +52,117 @@ impl ValueBufs {
...
+            Self::Normal { .. } => None,
+            Self::SharedAnchor { shared_index, .. } | Self::SharedDerived { shared_index } => {
+                Some(*shared_index)
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum FormulaRecord {
```
jmcnamara: You should  add some comment here about what this structure is and why it is a similar private copy of `XlsxFormulaMetadata`. I think I see why it is used but future readers of the code (including me) might not.

Also, same comment as before about the variant name.

Also, maybe the name should be `FormulaMetadata` like `XlsxFormulaMetadata`. Either way, `FormulaRecord` or Record in general, isn't very informative.

src/xlsx/cells_reader.rs
```diff
@@ -49,6 +52,117 @@ impl ValueBufs {
...
+            Self::SharedDerived { shared_index, .. } => {
+                XlsxFormulaMetadata::SharedDerived { shared_index }
+            }
+        }
+    }
+}
+
+/// A single XLSX cell record containing both cached/literal value and expanded formula text.
+#[derive(Clone, Debug, PartialEq)]
+pub struct XlsxCellFormulaRecord<'a> {
```
jmcnamara: I'd suggest  the `XlsxCellFormula` name for this structure. Also, then in the comment "A single XLSX cell structure ...".

src/xlsx/cells_reader.rs
```diff
@@ -49,6 +52,117 @@ impl ValueBufs {
...
+pub enum XlsxFormulaMetadata {
+    /// Ordinary, non-shared formula.
+    Normal {
+        /// Formula text.
+        formula: String,
+    },
+    /// Shared formula anchor/template cell.
+    SharedAnchor {
+        /// Shared formula index (`si`). Shared formula indices are worksheet-local.
+        shared_index: usize,
```
jmcnamara, line 68: Add one blank line between each member variant, here and elsewhere, when some/all have comments.

src/xlsx/cells_reader.rs
```diff
@@ -211,6 +328,204 @@ where
...
+                    return Err(XlsxError::Unexpected(
+                        "si attribute is mandatory if it is shared",
+                    ));
+                }
+            };
+
+            return match get_attribute(e.attributes(), QName(b"ref"))? {
+                Some(res) => {
+                    let reference = get_dimension(res)?;
+                    if let Some(f) = formula {
```
jmcnamara: I think this may always be `Some()` in this condition, in which case the `else` won't be reached. Could you double check.

src/xlsx/mod.rs
```diff
@@ -3000,10 +3067,213 @@ impl<'a, RS: Read + Seek + 'a> Iterator for PivotCacheIter<'a, RS> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::io::Write;
+    use std::io::{Cursor, Write};
     use zip::write::SimpleFileOptions;
     use zip::ZipWriter;
 
+    fn minimal_xlsx(sheet_xml: &str, shared_strings_xml: Option<&str>) -> Vec<u8> {
```
jmcnamara: Same comment as before. Remove this inline file creation and replace it with one of the supplied actual files or move this to an integration test.

jmcnamara (changes requested):

jmcnamara:
> Just checking in on this PR.

Apologies. I haven't had time to do reviews recently. I had started this one and I had some notes. Let me get to it now.

tests/test.rs
```diff
@@ -1942,6 +1942,99 @@ fn issue_391_shared_formula() {
...
+            formula_count += 1;
+            last_formula = record.formula;
+        }
+    }
+
+    assert_eq!(formula_count, 100);
+    assert_eq!(last_formula, Some("A100*2".to_string()));
+}
+
 #[test]
```
jmcnamara, line 2040: Can you also ensure that the non_monotonic_si_shared_formula() and associated file passes with the new streaming interface.

jmcnamara:
Overall comment, it would be good to add an example of reading formulas and values with this new streaming method. Demonstrate how read a short number of typical formulas. It shouldn't focus on the shared formula part since that (hopefully) will be transparent to end users.

jmcnamara:
Not directed at this PR but `XlsxCellReader `now has four methods that all advance `self.xm`: `next_cell`, `next_formula`, `next_cell_with_formula`, `next_cell_with_formula_metadata`  and mixing them on the same reader silently skips cells. I don't have a good solution for this (off the top of my head) but this isn't going to scale. I think #659 is adding another.

PSU3D0:
> Not directed at this PR but `XlsxCellReader `now has four methods that all advance `self.xm`: `next_cell`, `next_formula`, `next_cell_with_formula`, `next_cell_with_formula_metadata` and mixing them on the same reader silently skips cells. I don't have a good solution for this (off the top of my head) but this isn't going to scale. I think #659 is adding another.

I think a builder type pattern with either typestates or Option<T> unified struct (with projection to existing `next_` methods) makes some sense but it's definitely more of a design decision.

```rust
let mut reader = workbook
    .worksheet_stream("Sheet1")
    .values(true)
    .formulas(FormulaMode::Expanded)
    .build()?;
```

Then the reader only exposes:

```rust
reader.next()
```

Formula modes could be:

```rust
FormulaMode::None
FormulaMode::Expanded
FormulaMode::Metadata
```

If users are expecting to mix-and-match then:

```rust
XlsxCellRecord {
  pos,
  value,
  formula: Option<_>,
  hyperlink: Option<_>,
  style: Option<_>,
}
```

or maybe something on `Stream<Item = <...>>` where:

1. values only
```text
Item = Cell<DataRef>
```
2. values + expanded formula text
```text
Item = XlsxCellFormula
```
3. values + formula metadata
```text
Item = XlsxCellFormulaMetadataRecord
```
4. Maybe formula-only
```text
Item = Cell<String>

PSU3D0:
> Overall comment, it would be good to add an example of reading formulas and values with this new streaming method. Demonstrate how read a short number of typical formulas. It shouldn't focus on the shared formula part since that (hopefully) will be transparent to end users.

Should be addressed in [1f42dc6](https://github.com/tafia/calamine/pull/654/commits/1f42dc6cc998c4e4a0727928688a4d1fa436987a)

jmcnamara:
Thanks. Remember to resolve the comments that have been resolved so that I know they have been addressed. And then let me know the PR is ready for merge.

PSU3D0:
> Thanks. Remember to resolve the comments that have been resolved so that I know they have been addressed. And then let me know the PR is ready for merge.

Resolved. Should be ready for mege :racing_car:

jmcnamara:
Merged. Thanks.

PR #656, opened 2026-05-14
xlsx: add support for rstring records

### LINKED ISSUE #655: Bug: Unable to retrieve xls cell values using RSTRING records
### Current behavior

Currently running the sample code produces this output:

```
Got value from cell: None
```

### Expected behavior

It should be able to read the cell value:

```
Got value from cell: Some(String("ELECTORAL DIVISION OF SOMERSET EAST ; FIELD-CÓRNËTCY No. 5, ZWAGERSHÓÊN."))
```

### Sample code to reproduce

```rust
use calamine::{
    Error, Reader, Xls, open_workbook,
};

    let test_file = "biff5-rich-text-string.xls";
    let mut workbook: Xls<_> = open_workbook(test_file)?;
    let sheet_range = workbook.worksheet_range("Sheet1")?;
    println!("Got value from cell: {:?}", sheet_range.get_value((0, 0)));
    Ok(())
}
```

### Test file

This file is from the test suite of the [R readxl package](https://github.com/tidyverse/readxl). It displays fine in both LibreOffice & MS Excel.

[biff5-rich-text-string.xls](https://github.com/user-attachments/files/27744240/biff5-rich-text-string.xls)

It was apparently produced by ABBYY FineReader (OCR software).

### Environment

```text
- calamine version: 0.35
- Cargo.toml dependency line for calamine: calamine = "0.35.0"
- rustc version: rustc 1.94.0
- OS: windows/linux
```

### Checklist

aquasync:
Closes #655


aquasync:
For the original issue in the R package see https://github.com/tidyverse/readxl/issues/611.

It was fixed in the underlying C library https://github.com/libxls/libxls/commit/b6d9d872756f69780b743dbaec9cd2ec30c37740.

Approach here is the same - read using the label code path.

Issue #663, opened 2026-06-16
expand_shared_formula_with_offset_into is not greedy enough for named ranges in xlsx


OSjoerdWie:
### Current behavior

When supplying the name `NO1_VARIABLE` within a formula, `NO1` will be recognized as a valid cell reference and updated accordingly within the formula (depending on the direction, either `NP1` or `NO2`).

### Expected behavior

`NO1_VARIABLE` should remain after updating the shared formula for a cell. 

### Sample code to reproduce

```rust
use calamine::{open_workbook, Error, Reader, Xlsx};

fn main() -> Result<(), Error> {
    let test_file = "somefile.xlsx";

    let mut workbook: Xlsx<_> = open_workbook(test_file)?;

    let sheet_range = workbook.worksheet_formula("Sheet1")?;

    let mut i = 0; 
    let columns = vec!["A", "B", "C", "D", "E", "F"];

    sheet_range.range((6, 0), (6,5)).cells().for_each(|(_, _, formula)| {
        assert_eq!(*formula, format!("{}6+NO1_VARIABLE", columns[i]));
        i += 1;
    });   
    
    Ok(())
}
```

### Test file

[somefile.xlsx](https://github.com/user-attachments/files/29016531/somefile.xlsx)

### Environment

```text
- calamine version: 0.35.0
- Cargo.toml dependency line for calamine: calamine = "0.35.0"
- rustc version: 1.94.0
- Excel/OpenOffice/LibreOffice version: Microsoft Excel 365 Apps for Enterprise Version 2607 (Beta)
- OS: Windows 11
```

### Checklist

- [x] I have added a complete sample program that compiles in Rust.
- [x] I have added a test file.

jmcnamara:
Thanks for the detailed report. I'll review the fix.

jmcnamara:
Closed by #664

PR #664, opened 2026-06-16
Make shared formula cell replacement more greedy

OSjoerdWie:
Fixes the issue described in #663 

Note that I made an additional (non issue related) commit due to build failure of the forked repo.

### DIFF
--- a/src/xlsx/mod.rs
+++ b/src/xlsx/mod.rs
@@ -3460,7 +3460,14 @@ fn expand_shared_formula_with_offset_into(
     let mut token_end = 0;
 
     for (i, &c) in bytes.iter().enumerate() {
-        if !in_quote && (c.is_ascii_alphanumeric() || c == b'$' || c == b':') {
+        if !in_quote
+            && (c.is_ascii_alphanumeric()
+                || c == b'\\'
+                || c == b'.'
+                || c == b'_'
+                || c == b'$'
+                || c == b':')
+        {
             token_end = i + 1;
         } else {
             if token_start < token_end {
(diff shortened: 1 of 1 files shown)


examples/read_picture_data.rs
```diff
@@ -26,8 +26,10 @@
 //!   Sheet: "Sheet1", Cell: (2, 0), Type: png, Size: 178 bytes
 //! ```
 
+#[cfg(feature = "picture")]
```
jmcnamara: Remove these `#[cfg(feature = "picture")]` changes. They don't relate to the shared formula bug.

Preferably rebase and remove the "picture" commit and force push the single fix commit to your branch. The PR will get updated after that.

src/xlsx/mod.rs
```diff
@@ -4616,6 +4616,31 @@ mod tests {
             "한글 B1 テスト".to_owned()
         );
 
+        assert_eq!(
```
jmcnamara, line 4626: Here are some other relevant tests you could add.

```rust
  // Named range after a cell ref.
  assert_eq!(replace_cell_names("A6+NO1_VARIABLE", (0, 1)).unwrap(), "B6+NO1_VARIABLE");
  assert_eq!(replace_cell_names("A6+NO1_VARIABLE", (1, 0)).unwrap(), "A7+NO1_VARIABLE");

  // Named range before a cell ref.
  assert_eq!(replace_cell_names("NO1_VARIABLE+Q5", (0, 1)).unwrap(), "NO1_VARIABLE+R5");

  // Leading underscore named range — should pass through.
  assert_eq!(replace_cell_names("_PRIVATE+A1", (0, 1)).unwrap(), "_PRIVATE+B1");

  // Named range with multiple underscores.
  assert_eq!(replace_cell_names("NO1__DOUBLE+A1", (0, 1)).unwrap(), "NO1__DOUBLE+B1");
```

jmcnamara (changes requested):

jmcnamara:
The core change looks good. I have made some change requests to other parts of the PR.

@louisdewar can you double check this. The change came in with your commit 38ef0e38


> Note that I made an additional (non issue related) commit due to build failure of the forked repo.

I'm not sure why that is. It didn't fail on main. I'll look into it. But either way, omit it from this PR.

jmcnamara:
Could you:

1. Mark the comments as resolved, if they are resolved.
2. Fix the formatting issue.: ` cargo fmt -- --check`
3. Squash the update into the main commit.

louisdewar:
Hi, yes I wasn't thinking about defined names with this.
This fix seems generally good, as far as I can tell the logic is to expand the definition of what is considered to "possibly" be a range to include characters that constitute defined names and then let the defined name get rejected later on when it's seen that it's not really a valid range. Before it would tokenise <NO1> and think that was a valid range, now it would tokenise <NO1_VARIABLE> and reject it as a range to be "replaced".

I think there's a few issues with some edge cases:
1. NO1.VARIABLE
2. \A1

I believe these are all "valid" defined names which I think will get treated as ranges.

I _think_ you can fix by just adding '.' and '\\' to the char compare.

jmcnamara:
@OSjoerdWie Could you add test cases and a fix for the two cases @louisdewar lists above `NO1.VARIABLE` and
`\A1`.

OSjoerdWie:
Will try to do so this week.

jmcnamara (approved):

jmcnamara:
Merged. Thanks.

PR #665, opened 2026-06-19
tests(xlsx): add regression for date cells with quoted custom number formats

jnsn99:
## Summary

Adds a regression test exercising date cells whose `numFmt formatCode` contains quoted literal text (e.g. `mmmm" "yyyy`).

## Background

Such formats are stored in `xl/styles.xml` with `&quot;` XML entity references. Prior to #621, `read_styles` UTF-8 decoded the raw `Attribute::value` bytes without unescaping and handed the result to `detect_custom_number_format`. The `;` arm of that function then returned `CellFormat::Other` because of the `;` inside `&quot;`, so the date cell was reported as a plain `Float(serial)` instead of `DateTime`.

#621 (`perf: custom zero-overhead attribute extraction`) rewrote the attribute extraction path through `get_attrs!` + `decode_attr`, which now unescapes entity references. That incidentally fixed the bug, but no test was added to guard against future regressions.

### Before/after #621

For a cell holding serial `44197.0` styled with a quoted custom format(e.g. `mmmm" "yyyy`), `range.get_value(...)` returned:

```
// v0.35.0
Some(Float(44197.0))

// master (after #621 / commit 5c4a52b)
Some(DateTime(ExcelDateTime { value: 44197.0, datetime_type: DateTime, is_1904: false }))
```

(Reproduced locally by running the test added in this PR against both refs: it passes on master HEAD and fails on `v0.35.0` with exactly the `Float` value above.)

## What this PR adds

- `tests/date_quoted_format.xlsx`: 4 cells with the same date serial (2021-01-01) but different quoted custom number formats:
  - `mmmm" "yyyy`
  - `yyyy"-"mm"-"dd`
  - `[$-409]mmmm" "d", "yyyy`
  - `[$-411]ggge"年"m"月"d"日";@` (Japanese era)
- `tests/test.rs::date_xlsx_custom_format_with_quoted_text`: asserts that each cell is read as `Data::DateTime`.

## Verification

- Passes on current `master` (after #621).
- Fails on `v0.35.0` with `Some(Float(44197.0))` vs the expected
  `Some(DateTime(...))`, confirming the test captures the prior bug.


jmcnamara:
Merged. Thanks.

PR #667, opened 2026-06-26
Handle the `*` fill operator in custom number format detection

mahirhir:
`detect_custom_number_format` (`src/formats.rs`) — the number-format classifier shared by xlsx, xlsb and xls — treats `\` and `_` as "escape the next char" but does not handle Excel's `*` fill operator. With `*c`, the character after `*` is a literal used to pad the cell, so a numeric format whose fill char is a date letter, e.g. `#,##0*y`, is misread as `DateTime` and a number cell is reported as a date.

This sits in the same `match` that #665 just hardened for quote handling; `*` was the remaining unhandled case. The fix adds `'*'` to the existing escape arm.

Verified both ways: `#,##0*y` returns `DateTime` before the fix and `Other` after; `*-yyyy-mm-dd` still correctly returns `DateTime`. All 24 existing format tests are unchanged and `cargo test --lib` passes 42/42.

### DIFF
--- a/src/formats.rs
+++ b/src/formats.rs
@@ -22,7 +22,12 @@ pub fn detect_custom_number_format(format: &str) -> CellFormat {
     for s in format.chars() {
         match (s, escaped, is_quote, ap, brackets) {
             (_, true, ..) => escaped = false, // if escaped, ignore
-            ('_' | '\\', ..) => escaped = true,
+            // `\` escapes, `_` skips a character's width, and `*` repeats the
+            // next character to fill the cell. In all three cases the following
+            // character is a literal, not a format token, so it must be ignored
+            // (otherwise a fill/skip char such as `d`, `m` or `y` is mistaken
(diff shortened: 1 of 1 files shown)


src/formats.rs
```diff
@@ -22,7 +22,12 @@ pub fn detect_custom_number_format(format: &str) -> CellFormat {
     for s in format.chars() {
         match (s, escaped, is_quote, ap, brackets) {
             (_, true, ..) => escaped = false, // if escaped, ignore
-            ('_' | '\\', ..) => escaped = true,
+            // `\` escapes, `_` skips a character's width, and `*` repeats the
```
jmcnamara: Reduce this down to a single line instructive comment.

jmcnamara (changes requested):

mahirhir:
Trimmed to a single line as suggested.

jmcnamara:
Merged. Thanks.

PR #668, opened 2026-06-26
fix: strip whitespace from <t> elements without xml:space="preserve"

dayongxie:
Per the XML spec, whitespace in element content should be normalized unless xml:space="preserve" is set. calamine's read_string_with_bufs previously preserved all whitespace unconditionally, causing cells with leading/trailing spaces, tabs, or newlines to return those characters.

This fix checks for xml:space="preserve" on <t> elements in shared strings and inline strings. When absent, leading/trailing whitespace (\t, \n, space, \r) is trimmed — but only if the string contains non-whitespace content, to preserve cells whose entire value is intentionally whitespace.

Covers both shared strings (read_shared_strings) and inline strings (cells_reader.rs), since both call read_string_with_bufs.


jmcnamara:
Thanks. Could you add some tests as part of the PR.

dayongxie:
I’ve added unit tests and also fixed an issue in the previous implementation.
I switched to using trim_text for whitespace normalization, which avoids inconsistencies from handling different paths separately.
Now both shared strings and inline strings follow the same logic, and the behavior matches the expected XML semantics.
If you have a moment, could you please take another look at this revision? 🙏

jmcnamara:
Some questions about what is being fixed:

1. What was the old behaviour, with an example.
2. What is the current behaviour, with the same example.
3. What commit changed the behaviour.

PR #669, opened 2026-06-27
fix(xlsx,ods): return Data::Int for whole-number worksheet cells

yertto:
Worksheet cell values without a decimal point (e.g. <v>95</v>) were always parsed as Data::Float, even though the xlsx format stores them as integers. Check the raw value for '.' and return Data::Int when absent and no date/time format applies.

This matches the existing behavior of the xlsb reader and PivotCache parser.

Assisted-by: OpenCode v1.17.9 (opencode/big-pickle)

### DIFF
--- a/src/ods.rs
+++ b/src/ods.rs
@@ -674,10 +674,20 @@ where
         match key {
             b"office:value" if !is_value_set => {
                 let v = reader.decoder().decode(value)?;
-                val = Data::Float(
-                    fast_float2::parse(v.as_bytes())
-                        .map_err(|_| OdsError::ParseFloat(v.parse::<f64>().unwrap_err()))?,
-                );
+                val = if !value.contains(&b'.') {
+                    match atoi_simd::parse::<i64>(v.as_bytes()) {
+                        Ok(n) => Data::Int(n),
+                        Err(_) => Data::Float(
+                            fast_float2::parse(v.as_bytes())
+                                .map_err(|_| OdsError::ParseFloat(v.parse::<f64>().unwrap_err()))?,
+                        ),
+                    }
+                } else {
+                    Data::Float(
+                        fast_float2::parse(v.as_bytes())
+                            .map_err(|_| OdsError::ParseFloat(v.parse::<f64>().unwrap_err()))?,
+                    )
+                };
                 is_value_set = true;
             }
             b"office:string-value" | b"office:date-value" | b"office:time-value"
--- a/src/xlsx/cells_reader.rs
+++ b/src/xlsx/cells_reader.rs
@@ -17,6 +17,9 @@ use crate::{
     Cell, XlsxError,
 };
 
+#[cfg(test)]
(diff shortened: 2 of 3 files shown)


jmcnamara:
> even though the xlsx format stores them as integers.

I don't think that is technically correct. 

Excel (I don't know about ODS) handles all numbers as 64bit IEEE754 Doubles. It stores them to XML/text with a C/C++ format like `"%.16G"` so `1.0` is written as `1` but it is still implicitly a float. You can also see this in the way that it doesn't store integers above 999,999,999,999,999 (15 digits, therefore f64 and not i64).

So for me this is a won't/don't fix.

yertto:
Hi @jmcnamara , was just seeing if it was possible to fix a downstream issue in nushell.

Before fix:
```
❯ nu -c "open tests/fixtures/formats/sample_data.xlsx | get SalesOrders | first"
╭───────────┬─────────────╮
│ OrderDate │ 8 years ago │
│ Region    │ East        │
│ Rep       │ Jones       │
│ Item      │ Pencil      │
│ Units     │ 95.00       │
│ Unit Cost │ 1.99        │
│ Total     │ 189.05      │
╰───────────┴─────────────╯
```

After fix:
```
❯ ./target/debug/nu -c "open tests/fixtures/formats/sample_data.xlsx | get SalesOrders | first"
╭───────────┬─────────────╮
│ OrderDate │ 8 years ago │
│ Region    │ East        │
│ Rep       │ Jones       │
│ Item      │ Pencil      │
│ Units     │ 95          │
│ Unit Cost │ 1.99        │
│ Total     │ 189.05      │
╰───────────┴─────────────╯
```

ie. where `Units` is correctly shown as `95` instead of `95.00`.

ie. using [tests/fixtures/formats/sample_data.xlsx](https://github.com/nushell/nushell/blob/main/tests/fixtures/formats/sample_data.xlsx)

yertto:
> Excel (I don't know about ODS) handles all numbers as 64bit IEEE754 Doubles. It stores them to XML/text with a C/C++ format like "%.16G" so 1.0 is written as 1 but it is still implicitly a float. 

Thanks @jmcnamara , I'll just "fix" it downstream instead.

Closed by https://github.com/nushell/nushell/pull/18485

jmcnamara:
> I'll just "fix" it downstream instead.

That is the probably the best place to fix it. Note however, that if you have a list of floating point numbers where some have no decimals then they will be formatted differently to the others. For example `[1.25, 2.00, 3.54, 4.00, 5.85]` may appear as the following with your fix:

```
1.25
2
3.54
4
5.85
```
This is why tools like Polars work out the  type of the entire column in a dataframe before coercing the type. I don't know if nushell does that.

yertto:
Yer, you're not wrong.
Forcing Integers for some Floats may cause issues for some folks.

Interesting about the Polars solution, but I might leave it for now, as I'm thinking that avoiding forcing Floats for ALL Integers is a "good enough" fix.

PR #670, opened 2026-06-30
More informative error if the file is very short

sftse:
If the file is very short, the error path will be triggered on an `fn read_exact` call, leaving an uninformative IO error, although it is probably better to hint that the file is invalid xls.

### DIFF
--- a/src/cfb.rs
+++ b/src/cfb.rs
@@ -49,11 +49,13 @@ impl std::fmt::Display for CfbError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             CfbError::Io(e) => write!(f, "I/O error: {e}"),
-            CfbError::Ole { len, magic } if *len < 512 => write!(
-                f,
-                "Invalid CFB file ({len} bytes is too small), magic: {:x?}",
-                &magic[..8.min(*len)]
-            ),
+            CfbError::Ole { len, magic } if *len < 512 => {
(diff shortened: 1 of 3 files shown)


src/cfb.rs
```diff
@@ -46,7 +49,15 @@ impl std::fmt::Display for CfbError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             CfbError::Io(e) => write!(f, "I/O error: {e}"),
-            CfbError::Ole => write!(f, "Invalid OLE signature (not an office document?)"),
+            CfbError::Ole { len, magic } if *len < 512 => write!(
+                f,
+                "Invalid CFB file ({len} bytes is too small), magic: {:x?}",
+                &magic[..8.min(*len)]
```
jmcnamara: Is this a useful error message? Most users won't know what `magic` means here. The probably won't know what CFB is either and it isn't mentioned in the other error messages. Maybe something like`Invalid file. OLE stream is too small: {len} bytes` (or similar).

sftse: `Cfb error: Invalid CFB file (xxx bytes is too small)` is how it would appear to the user, the cfb prefix gets added when printing `XlsError` but I agree this could be better.

sftse: Pushed an alternative suggestion that tries to combine the terminology previously here with the information when the file is too short and file signature mismatch.

jmcnamara:
Merged. Thanks.

PR #671, opened 2026-07-02
fix: strip whitespace from <t> elements without xml:space="preserve"

dayongxie:
## Summary

Per the XML spec, whitespace in element content should be normalized unless xml:space="preserve" is set. calamine's read_string_with_bufs previously preserved all whitespace unconditionally, causing cells with leading/trailing spaces, tabs, or newlines to return those characters.

This fix checks for xml:space="preserve" on <t> elements in shared strings and inline strings. When absent, quick-xml's trim_text(true) is enabled to strip leading/trailing whitespace from the text content.

Covers both shared strings (read_shared_strings) and inline strings (cells_reader.rs), since both call read_string_with_bufs.

## Behavior examples

| Raw XML | Old behavior | New behavior |
|---|---|---|
| `<t>  trimmed value  </t>` | `"  trimmed value  "` | `"trimmed value"` |
| `<t>\n\tvalue with tabs\n</t>` | `"\n\tvalue with tabs\n"` | `"value with tabs"` |
| `<t xml:space="preserve">  kept  </t>` | `"  kept  "` | `"  kept  "` (unchanged) |
| `<t> \t\n</t>` (whitespace-only) | `" \t\n"` | `""` (empty) |
| `<t>  line1\nline2  </t>` | `"  line1\nline2  "` | `"line1\nline2"` (inner newline preserved) |

## On the behavior change

This is not a regression from a specific commit — calamine has always preserved whitespace as-is since the initial implementation. The current behavior is inconsistent with the XML specification (which requires whitespace normalization by default) and with how Excel and other spreadsheet tools handle text content.

Supersedes #668 (force-push squash closed the original).


dayongxie:
@jmcnamara, thanks for your review on #668. To answer your questions:

1. **Old behaviour**: Whitespace was preserved as-is. For example, `<t>  trimmed value  </t>` would return `"  trimmed value  "` (with leading/trailing spaces), and `<t>\n\tvalue\n</t>` would return the string with all newlines and tabs intact.

2. **New behaviour**: Whitespace is trimmed via `quick-xml`'s `trim_text(true)` when `xml:space="preserve"` is not set on the `<t>` element. The same `<t>  trimmed value  </t>` now returns `"trimmed value"`. When `xml:space="preserve"` is present, whitespace is preserved as before.

3. **Which commit changed the behaviour**: This is not a regression from a specific commit — calamine has always preserved whitespace unconditionally since the initial implementation of `read_string_with_bufs`. The current behavior is inconsistent with the XML specification, which requires whitespace normalization unless `xml:space="preserve"` is explicitly set.

I've also updated the PR description with a table of behavior examples. (Force-push to squash commits closed the original PR, so I've recreated it here.)

jmcnamara:
Thanks for the answers. One more question, why are you fixing this? I understand that this makes the parsing consistent with the XML spec but why are **you** fixing it. Have you encountered real world cases where whitespace should have been stripped but isn't, or some other reason?

dayongxie:
We encountered this with real-world xlsx files in our data pipeline. Some files (generated by third-party tools) contain text nodes with leading/trailing whitespace. Excel displays these cells with the whitespace trimmed, but calamine reads them as-is, so the parsed values do not match what users see in Excel. This caused concrete data quality issues for us, which is what motivated the fix.

jmcnamara:
Thanks for the information and patch. Merged.

PR #675, opened 2026-07-05
fix(xlsb): consume unmatched workbook-global record bodies

### LINKED ISSUE #666: xlsb: read_excel returns 0 sheets (and sometimes panics) when a workbook-global record's body size is 144 or 156
### Current behavior

For some valid `.xlsb` workbooks, `read_excel(...).sheet_names` returns an **empty list** even though the workbook contains valid `BrtBundleSh` sheet records. A slightly different record size makes it **panic** instead (`index out of bounds`).

**Root cause:** in `src/xlsb/mod.rs`, `read_workbook` matches a few record ids and lets everything else fall through `_ => ()` **without consuming the record body**:

```rust
match iter.read_type()? {
    0x0099 => { /* WbProp */ }
    0x009C => { /* BrtBundleSh -> push sheet */ }
    0x0090 => break,        // BrtEndBundleShs
    _ => (),                // body NOT consumed
}
```

Because an unmatched workbook-global record's body is left unconsumed, the next `read_type()` reads that record's **size-varint bytes** as the next record id:

- body size **144** → size varint `90 01` → `0x90` read as `BrtEndBundleShs` → loop `break`s **before** any `BrtBundleSh` → **0 sheets**.
- body size **156** → size varint `9C 01` → `0x9C` read as `BrtBundleSh` → decoded from an empty buffer → **panic** (`range start index 8 out of range for slice of length 0`, mod.rs:325).
- sizes 143 (`8F 01`) and 145 (`91 01`) parse fine → it is the specific byte value, not the magnitude.

**Suggested fix:** consume the body for unmatched records, e.g. `_ => { let _ = iter.fill_buffer(&mut buf)?; }`.

### Expected behavior

All sheets are enumerated. The workbook is valid and its `BrtBundleSh` records decode cleanly, so `sheet_names()` should return every sheet — and it must never panic on a valid file.

### Sample code to reproduce

```rust
Minimal reproduction (Python via `fastexcel`, which wraps calamine 0.34.0):


import io, zipfile, struct, fastexcel

def rec(rid, payload=b""):
    o = bytearray()
    if rid < 0x80: o.append(rid)
    else: o += bytes([(rid & 0x7F) | 0x80, (rid >> 7) & 0xFF])
    s = len(payload)
    while True:
        b = s & 0x7F; s >>= 7
        o.append(b | 0x80 if s else b)
        if not s: break
    return bytes(o) + payload

def wstr(s): return struct.pack("<I", len(s)) + s.encode("utf-16-le")

def workbook_bin(toxic_size):
    out  = rec(0x0083) + rec(0x0099, b"\x00"*4)        # BeginBook, WbProp
    if toxic_size is not None:
        out += rec(0x0817, b"\x00"*toxic_size)         # any unmatched global record
    out += rec(0x008F)                                 # BeginBundleShs
    out += rec(0x009C, struct.pack("<II",0,0)+wstr("rId1")+wstr("Sheet1"))
    out += rec(0x0090) + rec(0x0084)                   # EndBundleShs, EndBook
    return out

def build(toxic_size):
    sheet = rec(0x0081)+rec(0x0094,struct.pack("<IIII",0,0,0,0))+rec(0x0091)+rec(0x0092)+rec(0x0082)
    b = io.BytesIO()
    with zipfile.ZipFile(b, "w") as z:
        z.writestr("[Content_Types].xml", '<?xml version="1.0"?><Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types"><Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/><Default Extension="bin" ContentType="application/vnd.ms-excel.sheet.binary.macroEnabled.main"/><Override PartName="/xl/worksheets/sheet1.bin" ContentType="application/vnd.ms-excel.worksheet"/></Types>')
        z.writestr("_rels/.rels", '<?xml version="1.0"?><Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"><Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="xl/workbook.bin"/></Relationships>')
        z.writestr("xl/workbook.bin", workbook_bin(toxic_size))
        z.writestr("xl/_rels/workbook.bin.rels", '<?xml version="1.0"?><Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"><Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/worksheet" Target="worksheets/sheet1.bin"/></Relationships>')
        z.writestr("xl/worksheets/sheet1.bin", sheet)
    return b.getvalue()

print(fastexcel.read_excel(build(None)).sheet_names)   # ['Sheet1']  (control)
print(fastexcel.read_excel(build(144)).sheet_names)    # []         (bug)
print(fastexcel.read_excel(build(156)).sheet_names)    # panics
```

### Test file

Minimal reproduction (Python via `fastexcel`, which wraps calamine 0.34.0):

```python
import io, zipfile, struct, fastexcel

def rec(rid, payload=b""):
    o = bytearray()
    if rid < 0x80: o.append(rid)
    else: o += bytes([(rid & 0x7F) | 0x80, (rid >> 7) & 0xFF])
    s = len(payload)
    while True:
        b = s & 0x7F; s >>= 7
        o.append(b | 0x80 if s else b)
        if not s: break
    return bytes(o) + payload

def wstr(s): return struct.pack("<I", len(s)) + s.encode("utf-16-le")

def workbook_bin(toxic_size):
    out  = rec(0x0083) + rec(0x0099, b"\x00"*4)        # BeginBook, WbProp
    if toxic_size is not None:
        out += rec(0x0817, b"\x00"*toxic_size)         # any unmatched global record
    out += rec(0x008F)                                 # BeginBundleShs
    out += rec(0x009C, struct.pack("<II",0,0)+wstr("rId1")+wstr("Sheet1"))
    out += rec(0x0090) + rec(0x0084)                   # EndBundleShs, EndBook
    return out

def build(toxic_size):
    sheet = rec(0x0081)+rec(0x0094,struct.pack("<IIII",0,0,0,0))+rec(0x0091)+rec(0x0092)+rec(0x0082)
    b = io.BytesIO()
    with zipfile.ZipFile(b, "w") as z:
        z.writestr("[Content_Types].xml", '<?xml version="1.0"?><Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types"><Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/><Default Extension="bin" ContentType="application/vnd.ms-excel.sheet.binary.macroEnabled.main"/><Override PartName="/xl/worksheets/sheet1.bin" ContentType="application/vnd.ms-excel.worksheet"/></Types>')
        z.writestr("_rels/.rels", '<?xml version="1.0"?><Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"><Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="xl/workbook.bin"/></Relationships>')
        z.writestr("xl/workbook.bin", workbook_bin(toxic_size))
        z.writestr("xl/_rels/workbook.bin.rels", '<?xml version="1.0"?><Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"><Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/worksheet" Target="worksheets/sheet1.bin"/></Relationships>')
        z.writestr("xl/worksheets/sheet1.bin", sheet)
    return b.getvalue()

print(fastexcel.read_excel(build(None)).sheet_names)   # ['Sheet1']  (control)
print(fastexcel.read_excel(build(144)).sheet_names)    # []         (bug)
print(fastexcel.read_excel(build(156)).sheet_names)    # panics
```

### Environment

- calamine version: 0.34.0 (used transitively via the fastexcel 0.20.2 Python wheel, which bundles calamine 0.34.0 through PyO3)

- Cargo.toml dependency line for calamine: N/A — not consumed directly via Cargo; it ships inside the prebuilt fastexcel==0.20.2 wheel. Effective version is calamine = "0.34.0".

- rustc version: N/A — reproduced through the prebuilt fastexcel wheel (no local Rust toolchain; the wheel is built upstream).

- Excel/OpenOffice/LibreOffice version: affected real files are .xlsb exported from Microsoft Excel. The minimal reproducer above is hand-constructed bytes and needs no spreadsheet application.

- OS: Windows 11 (the bug is OS-independent — it is in the BIFF12 record framing in xlsb/mod.rs).

### Checklist

- [ ] I have added a complete sample program that compiles in Rust.

momomuchu:
Fixes #666.

read_workbook's record loop only consumes the body for BrtWbProp (0x0099)
and BrtBundleSh (0x009C) before checking for BrtEndBundleShs (0x0090).
Every other workbook-global record falls into `_ => ()` and its body is
never read.

Because read_type() only consumes the type bytes, the next call reads
the unconsumed record's size varint as if it were a new type. For most
sizes this just self-heals a few bytes later, but a size of 144 or 156
encodes as a varint that reads back exactly as 0x0090 or 0x009C. So:

- size 144 gets misread as BrtEndBundleShs, the loop breaks before any
  real sheet is seen, and sheet_names() silently returns nothing.
- size 156 gets misread as BrtBundleSh, decoded from the wrong offset,
  and panics (index out of bounds).

This isn't limited to hand-crafted files. BrtAbsPath15 (the absolute
save-path record) is variable length and already present, unconsumed,
in this repo's own real Excel-produced fixtures (tests/issues.xlsb has
it at size 68). Its size depends only on how long the folder path was
when the file was saved, so a plain resave from a deep enough path can
land on 144 or 156 with no crafting involved.

The fix mirrors the existing 0x0099/0x009C arms: consume the body via
fill_buffer for any unmatched record too.

Added regression tests building minimal in-memory xlsb workbooks with a
synthetic unmatched record at the toxic sizes, plus a control case.

Note: there is a second, structurally identical unconsumed-body pattern
in the BrtName loop just below this one, in the same function. I left it
out to keep this fix focused on #666; happy to send a follow-up.


src/xlsb/mod.rs
```diff
@@ -340,7 +340,13 @@ impl<RS: Read + Seek> Xlsb<RS> {
                     }
                 }
                 0x0090 => break, // BrtEndBundleShs
-                _ => (),
+                _ => {
+                    // Consume the body of any unmatched workbook-global record.
+                    // Leaving it unread desyncs the stream: the next `read_type()`
+                    // call would misread this record's trailing bytes as the next
+                    // record id (see #666).
+                    let _ = iter.fill_buffer(&mut buf)?;
```
jmcnamara: Simplify this comment and don't mention the bug report number in the code since that requires the reader to go look up the issue. Instead explain briefly what non-obvious behaviour is happening like:

```rust
// Consume the body of any unmatched workbook-global record
// to avoid reading the trailing bytes in the next read.
```

tests/test.rs
```diff
@@ -3550,6 +3550,125 @@ fn test_whitespace_trim_shared_strings() {
     );
 }
 
+/// Helpers to build a synthetic minimal `.xlsb` (BIFF12) `xl/workbook.bin` stream
```
jmcnamara: Don't build test files programmatically like this. It is better to have the xlsx/xlsb files in the tests directory so they can be examined or reused in other tests.

It also would be preferable to use files created by Excel. If the files are generated programmatically then there is no guarantee that what is being tested is valid. At a minimum the file should be opened and resaved in Excel.

tests/test.rs
```diff
@@ -3550,6 +3550,125 @@ fn test_whitespace_trim_shared_strings() {
...
+            br#"<?xml version="1.0"?><Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"><Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/worksheet" Target="worksheets/sheet1.bin"/></Relationships>"#,
+        )
+        .unwrap();
+
+        zip.finish().unwrap().into_inner()
+    }
+}
+
+#[test]
+fn issue_666_xlsb_control_reads_one_sheet() {
```
jmcnamara: Don't use the issue name in the test title (here and elsewhere). Again this forces someone to go look up what issue #666 is. Give the test a descriptive title, then add a comment that explains what it is testing. At the end of the comment you can add `See issue #666`, but the reader shouldn't have to look it up for the current context.

jmcnamara (changes requested):

momomuchu:
Addressed all three:

- Simplified the comment at `src/xlsb/mod.rs`, no issue ref in the code.
- Gave the tests descriptive names (`test_xlsb_unmatched_workbook_record_does_not_lose_sheet` / `_does_not_panic`) with a `// See issue #666` comment.
- Moved the test data out of the source into two fixture files under `tests/`, loaded with the usual `wb(...)` pattern, and removed the in-test byte builder.

On "produced by Excel": I don't have Excel, and LibreOffice here reads `.xlsb` but its export filter errors out, so I can't resave a real one. The two fixtures are minimal hand-built `.xlsb` files that reproduce both failure modes from #666 (both tests fail without the fix). Are those acceptable, or could you point me to / attach a real Excel-saved `.xlsb` with an unmatched workbook-global record so I can swap them in?

jmcnamara:
> On "produced by Excel": I don't have Excel, and LibreOffice here reads `.xlsb` but its export filter errors out, so I can't resave a real one.

My concern is that this PR is fixing an issue that doesn't happen in practice. It may happen in practice but that hasn't been demonstrated with a real file produced by Excel.

The original bug report at #666 says "affected real files are .xlsb exported from Microsoft Excel" but doesn't attach one of these file, even though that is a requirement for the bug report. I will ask again there.

momomuchu:
Thanks for following up on #666, makes sense to get it from the source. I'll hold off touching the fixtures until we hear back from the reporter.

The bug itself (unmatched record body not consumed in `read_workbook`) is real independent of which file triggers it, the fixtures just reproduce the two failure modes from the report. If a real file surfaces a case they don't cover I'll adjust.

Anything else you want changed in the meantime, or are we just waiting on #666 now?

jmcnamara:
> Anything else you want changed in the meantime, or are we just waiting on #666 now?

Nothing to be done here. The fix looks good I just need to be sure that it is required. If there is no feedback on #666 I will merge it for the next release.

jmcnamara:
Merged. Thanks.

PR #679, opened 2026-07-13
xls: preserve empty strings

### LINKED ISSUE #678: Bug: Empty strings in xls are treated as blank cells
### Current behavior

An cell with a zero-length string (eg just pressing ' in excel) is not blank (ISBLANK is FALSE). The xlsx reader will preserve a zero-length string (eg `<t/>`), but the xls reader is discarding them.

My test case code will produce:

```
Got value from cell: Some(String(""))
Got value from cell: None
```

### Expected behavior

It should preserve the empty string for xls, as it does for xlsx:

```
Got value from cell: Some(String(""))
Got value from cell: Some(String(""))
```

### Sample code to reproduce

```rust
use calamine::{
    Error, Reader, Xls, Xlsx, open_workbook,
};

    let test_file = "empty-string.xlsx";
    let mut workbook: Xlsx<_> = open_workbook(test_file)?;
    let sheet_range = workbook.worksheet_range("Sheet1")?;
    println!("Got value from cell: {:?}", sheet_range.get_value((0, 0)));

    let test_file = "empty-string.xls";
    let mut workbook: Xls<_> = open_workbook(test_file)?;
    let sheet_range = workbook.worksheet_range("Sheet1")?;
    println!("Got value from cell: {:?}", sheet_range.get_value((0, 0)));

    Ok(())
}
```

### Test file

[empty-string.xls](https://github.com/user-attachments/files/29962090/empty-string.xls)
[empty-string.xlsx](https://github.com/user-attachments/files/29962089/empty-string.xlsx)

### Environment

```text
- calamine version: 0.36
- Cargo.toml dependency line for calamine: calamine = "0.36.0"
- rustc version: 1.94.0
- OS: windows/linux
```

### Checklist

aquasync:
Closes #678


tests/test.rs
```diff
@@ -3580,3 +3580,10 @@ fn too_small_xls() {
         Ok(_) | Err(_) => panic!("Is expected to return CfbError::Ole(_) error"),
     }
 }
+
+#[test]
+fn xls_empty_string() {
```
jmcnamara, line 3592: Could you add a comment here about what is being tested and at the end add "See issue #XXX".

jmcnamara (changes requested):

tests/test.rs
aquasync: Sure thing, just added a comment.

jmcnamara:
Merged. Thanks.

PR #681, opened 2026-07-14
Add support for OOXML format

### LINKED ISSUE #680: Bug: Fails to open XLSX saved in Strict ISO/IEC 29500 format
### Current behavior

When opening a specific file I get an error that I didn't get prior to version 0.36 (I confirmed that it works on 0.35.0). The error is `"RelationshipNotFound"`.

### Expected behavior

Opening the file should not trigger an error.

### Sample code to reproduce

```rust
use calamine::{open_workbook, Error, Reader, Xlsx};

    let test_file = "TRAINING.xlsx";

    let workbook: Xlsx<_> = open_workbook(test_file)?;
}
```

### Test file

[TRAINING.xlsx](https://github.com/user-attachments/files/30026289/TRAINING.xlsx)

### Environment

```text
- calamine version: 0.36
- Cargo.toml dependency line for calamine: calamine = 0.36
- rustc version: 1.96.1
- OS: macos
```

### Checklist

saks:
fixes #680 

It appears that the issue is caused by changes from https://github.com/tafia/calamine/commit/f45e2487f6de04b0085eaba67da4234d19c53daf and likely happening because of exact match to a specific string ("http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument").
It looks like we can support different namespaces that represent the same type of relationship (`officeDocument`).


tests/test.rs
```diff
@@ -3580,3 +3580,8 @@ fn too_small_xls() {
         Ok(_) | Err(_) => panic!("Is expected to return CfbError::Ole(_) error"),
     }
 }
+
+#[test]
+fn issue_680() {
```
jmcnamara: Rename this test (and test file) to something that describes what is being tested (without having to look up what issue 680 is). Something like:

```rust
// Test for xlsx file with  strict ISO/IEC 29500 relationship paths.
// See issue #680.
#[test]
fn test_xlsx_strict_iso_paths() {
    let _: Xlsx<_> = wb("strict_iso_paths.xlsx");
}
```

jmcnamara (changes requested):

jmcnamara:
Merged. Thanks.

saks:
Thank you @jmcnamara! 

Quick question: is there an opportunity to release the next version any time soon? 0.36.0 is broken for me (because of the bug that I've addressed here) but I have to upgrade because of `quick-xml` [vulnerability](https://rustsec.org/advisories/RUSTSEC-2026-0194.html). Using github repo directly is not an option in the restricted env. 

Thank you in advance!

jmcnamara:
@saks 

I can do a 0.36.1 release but I won't be able to do it any earlier than Sunday night or Monday.

saks:
That would be awesome! Appreciate it!

jmcnamara:
Fixed in release v0.36.1.

PR #682, opened 2026-07-15
Add Xlsx cell style reading support

jmcnamara:
This is a smaller version of the PR to add XLSX Styles support to Calamine in #653.

Reviewing that PR was difficult because some recent changes to attribute handling meant that it didn't apply cleanly. It also contained changes for Rich Text and Worksheet Layout in a single commit. So in order to review it I simplified it down to the Style additions and rebased it as best I could while making the changes to the code that I would have requested.

I also added a second commit on top of it that reads the cell information and the style in one read, since that is the use case that most people will have:

> xlsx: add combined cell value and style id streaming
>    
>    Add `XlsxCellReader::next_cell_with_style_id()` which returns each cell
    value together with its style id in a single pass over the worksheet,
    avoiding a second read when both values and styles are needed.

I will submit another PR on top of this that gives a more comprehensive Cell reader interface.


ddimaria:
Looks good @jmcnamara

PR #685, opened 2026-07-22
feat: parse workbook properties (core + extended)

### LINKED ISSUE #677: feature request: Parse Workbook Properties
### Feature Request

see: https://github.com/ToucanToco/fastexcel/issues/495

> Would it be possible to parse Workbook Properties? I am not sure if they are inside of https://docs.rs/calamine/latest/calamine/struct.Xlsx.html#method.metadata?
> 
> e.g. I usually do something like this with openpyxl and it would be nice to do it with fastexcel instead.
> ```py
> wb = load_workbook(...)
> author = wb.properties.creator
> user = wb.properties.lastModifiedBy
> ```

Alex-ley-scrub:
Closes #677.

This PR adds support for reading workbook document properties from XLSX and XLSB files:

- Core properties from `docProps/core.xml`:
  - `creator`
  - `last_modified_by`
  - `created`
  - `modified`
  - `title`
  - `subject`
  - `description`
  - `keywords`
  - `category`
  - `content_status`
  - `revision`
  - `version`

- Extended properties from `docProps/app.xml`:
  - `application`
  - `app_version`
  - `company`
  - `template`
  - `manager`

Usage:

```rust
use calamine::{open_workbook_auto};

let mut excel = open_workbook_auto(path)?;
let props = excel.workbook_properties()?;
println!("creator: {:?}", props.creator);
println!("last modified by: {:?}", props.last_modified_by);
```

If a file does not contain `docProps/core.xml` or `docProps/app.xml`, all properties remain `None` and parsing continues normally.

Tests and fixtures covering both populated and missing properties are included.


src/lib.rs
```diff
@@ -226,6 +226,62 @@ pub struct Metadata {
...
+/// (`docProps/core.xml`) and extended file properties
+/// (`docProps/app.xml`) found in XLSX packages.
+///
+/// Most fields are optional because they depend on the file format and
+/// whether the producing application wrote them.
+#[derive(Debug, Default, Clone, PartialEq)]
+#[non_exhaustive]
+pub struct WorkbookProperties {
+    /// Core property: creator (dc:creator).
+    pub creator: Option<String>,
```
jmcnamara, line 242: Could you put one blank line between each field for readability.

src/xlsx/mod.rs
```diff
@@ -4582,3 +4591,153 @@ mod tests {
...
+            Ok(Event::Eof) => break,
+            Err(e) => return Err(XlsxError::Xml(e)),
+            _ => (),
+        }
+    }
+    Ok(())
+}
+
+#[derive(Clone, Copy)]
+enum PropField {
```
jmcnamara: `PropField` isn't a very descriptive name. I'd suggest `DocProperty` or similar.

jmcnamara (changes requested):

jmcnamara:
Overall this is a good (and useful) PR. I would have probably mapped some of the elements to different types but I think having everything as `Option<String>` is a good interface. So keep that as it is.

Could you also add an example to the examples directory in the format of some of the other examples. Also add an entry to `examples/Readme.md`.

There is also the possibility to have user defined custom fields but that can be left for another PR if it is requested. In that case the field could be `custom: HashMap<String, String>`.

tests/test.rs
```diff
@@ -3609,6 +3609,40 @@ fn test_xlsx_strict_iso_paths() {
...
+#[test]
+fn test_xlsx_workbook_properties_missing() {
+    let excel: Xlsx<_> = wb("workbook_properties_missing.xlsx");
+    let props = excel.metadata().workbook_properties();
+
+    assert!(props.creator.is_none());
+    assert!(props.last_modified_by.is_none());
+    assert!(props.application.is_none());
+    assert!(props.company.is_none());
+
```
jmcnamara, line 3646: There is a missing `}` here, probably as a result of a merge conflict fix.

Alex-ley-scrub:
thanks @jmcnamara

yeah I can do custom props as well potentially - i.e. `docProps/custom.xml` - I actually did a PR for that in openpyxl 6 years ago: https://foss.heptapod.net/openpyxl/openpyxl/-/merge_requests/384

there's a lot of prior art as well:
- https://github.com/exceljs/exceljs/issues/2109
- https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Writer/Xlsx/DocProps.php#L217-L241

I think we may as well add it in here too, whilst you are I are both in this context etc.

I'll do this, no worries:

> Could you also add an example to the examples directory in the format of some of the other examples. Also add an entry to examples/Readme.md.

src/lib.rs
Alex-ley-scrub: Done.

src/xlsx/mod.rs
Alex-ley-scrub: Renamed to `DocProperty`.

Alex-ley-scrub:
@jmcnamara for the `docProps/custom.xml`

We'd need to support:
- "vt:i4" - i.e. int like 4
- "vt:r8" - i.e. float/decimal like 3.142
- "vt:bool" - i.e. bool like true
- "vt:filetime" - i.e. datetime / iso timestamp like "2020-08-24T20:19:22Z"
- "vt:lpwstr" - i.e. string like "hello" but also can be self closing tag used with `<propety linkTarget="SomeName"><vt:lpwstr /></property>`

so would you still want the type as `custom: HashMap<String, String>` and just let the caller figure it out? Or we could return an enum with the VariantTypes info?

Alex-ley-scrub:
Added `docProps/custom.xml` support as discussed:

- New `CustomPropertyValue` enum covering `vt:i4`, `vt:r8`, `vt:bool`, `vt:filetime`, and `vt:lpwstr` (including `linkTarget` linked strings).
- Accessors for `as_i32`, `as_f64`, `as_bool`, `as_str`, `vt_type`, plus a `chrono`-gated `as_datetime` for file times.
- Added `examples/read_properties.rs` and updated `examples/Readme.md`.
- New fixture `tests/workbook_custom_properties.xlsx` with tests for all supported types.

`cargo test` and `cargo clippy --all-targets --examples` are green.

jmcnamara:
> https://foss.heptapod.net/openpyxl/openpyxl/-/merge_requests/384

That is good work. :-)



> for the `docProps/custom.xml` We'd need to support:

In `rust_xlsxwriter` I implemented a `CustomProperty` type for this:

https://github.com/jmcnamara/rust_xlsxwriter/blob/main/src/properties.rs#L613-L621

However that is probably overkill. Custom properties will not be a requirement for most users and anyone who did need them probably wouldn't be concerned with the type or could probably figure it out from the field name. Hence my suggestion to just use `HashMap<String, String>`.

However, even at that I don't think it is worth the effort of adding custom property support, at this time.

A better use of your/my time would be to add property support for the other file types: xlsb, xls, and ods. I never insist that features are rolled out across all file types but it seems important that users be able to access the author and creation/modification data, at a minimum.

Maybe you could look at xlsb. I can look at ODS (or vice-versa). @sftse would it be possible to extract basic property information like author and creation/modification from xls using the current cfb support?

Alex-ley-scrub:
Updated `examples/read_properties.rs` to match the conventions in the examples directory:

- Added a module-level rustdoc comment explaining what the example does and how to run it, including expected output.
- Removed the padded alignment in the `println!` calls.

`cargo run -q --example read_properties -- tests/workbook_custom_properties.xlsx` now prints:

```text
Core / Extended properties:
  creator: Test Creator
  application: Microsoft Excel

Custom properties:
  MyInt: 4 (vt:i4)
  MyBool: true (vt:bool)
  MyDateTime: 2020-08-24T20:19:22Z (vt:filetime)
  MyString: hello (vt:lpwstr)
  MyLink: SomeName (vt:lpwstr)
  MyFloat: 2.5 (vt:r8)
```

jmcnamara:
Thanks. Your additions for custom property handling came at more or less the same time as my comments about omitting it. If I had seen the work first I would have left that go. However, splitting them out into a separate PR is good too. It makes this review a bit simpler.

I'll review the current code but one high level comment. The current approach reads (or tries to read) the document properties for every xlsx/xlsb file (and later others) even if the user doesn't need them. That is a performance overhead, even if (relative to the data in the worksheets) it isn't a big one. A better approach might be to load the properties lazily when requested and cache them for repeated lookups (unlikely).

Something like this:

```diff
)$ git diff props-eager props-lazy src
diff --git a/src/lib.rs b/src/lib.rs
index fac7b06..40262de 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -226,8 +226,6 @@ pub struct Metadata {
     sheets: Vec<Sheet>,
     /// Map of sheet names/sheet path within zip archive
     names: Vec<(String, String)>,
-    /// Workbook properties (core and extended)
-    pub(crate) workbook_properties: WorkbookProperties,
 }
 
 /// Workbook document properties.
@@ -293,13 +291,6 @@ pub struct WorkbookProperties {
     pub manager: Option<String>,
 }
 
-impl Metadata {
-    /// Returns the workbook document properties.
-    pub fn workbook_properties(&self) -> &WorkbookProperties {
-        &self.workbook_properties
-    }
-}
-
 /// Type of sheet.
 ///
 /// Only Excel formats support this. Default value for ODS is
diff --git a/src/ods.rs b/src/ods.rs
index dd98f74..05a2304 100644
--- a/src/ods.rs
+++ b/src/ods.rs
@@ -206,7 +206,6 @@ where
         let metadata = Metadata {
             sheets: sheets_metadata,
             names: defined_names,
-            workbook_properties: crate::WorkbookProperties::default(),
         };
 
         Ok(Ods {
diff --git a/src/xlsb/mod.rs b/src/xlsb/mod.rs
index dfb85a2..6faa72a 100644
--- a/src/xlsb/mod.rs
+++ b/src/xlsb/mod.rs
@@ -162,6 +162,8 @@ pub struct Xlsb<RS> {
     formats: Vec<CellFormat>,
     is_1904: bool,
     metadata: Metadata,
+    /// Workbook document properties, parsed lazily on first access
+    workbook_properties: Option<Box<WorkbookProperties>>,
     #[cfg(feature = "picture")]
     pictures: Option<Vec<(String, Vec<u8>)>>,
     options: XlsbOptions,
@@ -440,14 +442,28 @@ impl<RS: Read + Seek> Xlsb<RS> {
         )
     }
 
-    fn read_properties(&mut self) -> Result<(), XlsbError> {
-        let mut props = WorkbookProperties::default();
-        read_core_properties(&mut self.zip, &mut props, &self.zip_path_cache)
-            .map_err(xlsx_error_to_xlsb)?;
-        read_app_properties(&mut self.zip, &mut props, &self.zip_path_cache)
-            .map_err(xlsx_error_to_xlsb)?;
-        self.metadata.workbook_properties = props;
-        Ok(())
+    /// Get the workbook document properties.
+    ///
+    /// Reads the core [`WorkbookProperties`] document properties, such as file
+    /// author and creation date/time. Missing fields are set as `None`.
+    ///
+    /// # Errors
+    ///
+    /// - [`XlsbError::Xml`] if a `docProps` part is present but malformed.
+    ///
+    pub fn workbook_properties(&mut self) -> Result<&WorkbookProperties, XlsbError> {
+        if self.workbook_properties.is_none() {
+            let mut props = WorkbookProperties::default();
+            read_core_properties(&mut self.zip, &mut props, &self.zip_path_cache)
+                .map_err(xlsx_error_to_xlsb)?;
+            read_app_properties(&mut self.zip, &mut props, &self.zip_path_cache)
+                .map_err(xlsx_error_to_xlsb)?;
+            self.workbook_properties = Some(Box::new(props));
+        }
+        Ok(self
+            .workbook_properties
+            .as_deref()
+            .expect("populated immediately above"))
     }
 
     #[cfg(feature = "picture")]
@@ -495,6 +511,7 @@ impl<RS: Read + Seek> Reader<RS> for Xlsb<RS> {
             formats: Vec::new(),
             is_1904: false,
             metadata: Metadata::default(),
+            workbook_properties: None,
             #[cfg(feature = "picture")]
             pictures: None,
             options: XlsbOptions::default(),
@@ -504,7 +521,6 @@ impl<RS: Read + Seek> Reader<RS> for Xlsb<RS> {
         xlsb.read_styles()?;
         let relationships = xlsb.read_relationships()?;
         xlsb.read_workbook(&relationships)?;
-        xlsb.read_properties()?;
         #[cfg(feature = "picture")]
         xlsb.read_pictures()?;
 
diff --git a/src/xlsx/mod.rs b/src/xlsx/mod.rs
index c2728ea..d87d12e 100644
--- a/src/xlsx/mod.rs
+++ b/src/xlsx/mod.rs
@@ -265,6 +265,8 @@ pub struct Xlsx<RS> {
     is_1904: bool,
     /// Metadata
     metadata: Metadata,
+    /// Workbook document properties, parsed lazily on first access.
+    workbook_properties: Option<Box<WorkbookProperties>>,
     /// Pictures
     #[cfg(feature = "picture")]
     pictures: Option<Vec<Picture>>,
@@ -547,12 +549,26 @@ impl<RS: Read + Seek> Xlsx<RS> {
         Ok(())
     }
 
-    fn read_properties(&mut self) -> Result<(), XlsxError> {
-        let mut core = WorkbookProperties::default();
-        read_core_properties(&mut self.zip, &mut core, &self.zip_path_cache)?;
-        read_app_properties(&mut self.zip, &mut core, &self.zip_path_cache)?;
-        self.metadata.workbook_properties = core;
-        Ok(())
+    /// Get the workbook document properties.
+    ///
+    /// Reads the core [`WorkbookProperties`] document properties, such as file
+    /// author and creation date/time. Missing fields are set as `None`.
+    ///
+    /// # Errors
+    ///
+    /// - [`XlsxError::Xml`] if a `docProps` part is present but malformed.
+    ///
+    pub fn workbook_properties(&mut self) -> Result<&WorkbookProperties, XlsxError> {
+        if self.workbook_properties.is_none() {
+            let mut props = WorkbookProperties::default();
+            read_core_properties(&mut self.zip, &mut props, &self.zip_path_cache)?;
+            read_app_properties(&mut self.zip, &mut props, &self.zip_path_cache)?;
+            self.workbook_properties = Some(Box::new(props));
+        }
+        Ok(self
+            .workbook_properties
+            .as_deref()
+            .expect("populated immediately above"))
     }
 
     fn read_relationships(&mut self) -> Result<HashMap<Vec<u8>, (String, String)>, XlsxError> {
@@ -2560,6 +2576,7 @@ impl<RS: Read + Seek> Reader<RS> for Xlsx<RS> {
             sheets: Vec::new(),
             tables: None,
             metadata: Metadata::default(),
+            workbook_properties: None,
             #[cfg(feature = "picture")]
             pictures: None,
             merged_regions: None,
@@ -2571,7 +2588,6 @@ impl<RS: Read + Seek> Reader<RS> for Xlsx<RS> {
         xlsx.read_styles()?;
         let relationships = xlsx.read_relationships()?;
         xlsx.read_workbook(&relationships)?;
-        xlsx.read_properties()?;
         #[cfg(feature = "picture")]
         xlsx.read_pictures()?;
 
@@ -4577,6 +4593,7 @@ mod tests {
             formats: vec![],
             is_1904: false,
             metadata: Metadata::default(),
+            workbook_properties: None,
             #[cfg(feature = "picture")]
             pictures: None,
             merged_regions: None,
```

src/lib.rs
```diff
@@ -226,6 +226,78 @@ pub struct Metadata {
     sheets: Vec<Sheet>,
     /// Map of sheet names/sheet path within zip archive
     names: Vec<(String, String)>,
+    /// Workbook properties (core and extended)
+    pub(crate) workbook_properties: WorkbookProperties,
+}
+
+/// Workbook document properties.
+///
+/// These properties correspond to the core document properties
```
jmcnamara: Since the document properties will be generic across all supported file types this comment shouldn't mention `docProps/core.xml` or `docProps/app.xml` (which won't be useful information for most users anyway). Instead say something like:

```rust
/// These properties correspond to the core document properties
/// like author and file creation date/time.
```

jmcnamara (changes requested):

Alex-ley-scrub:
I've addressed the lazy-loading feedback from https://github.com/tafia/calamine/pull/685#issuecomment-5146559861 @jmcnamara 

Workbook properties are now loaded lazily on first access (and cached) instead of being parsed eagerly during reader initialization, and the access pattern is now `workbook.workbook_properties()?` for the supported readers.

I also opened the deferred custom-properties follow-up as a stacked PR on my fork: https://github.com/Alex-ley-scrub/calamine/pull/1 - I can then open a PR from my fork into this repo once this PR here merges.
Tracking issue: https://github.com/tafia/calamine/issues/702

Want/need me to do anything else @jmcnamara?

examples/read_properties.rs
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright 2016-2026, Johann Tuffe.
+
+//! Demonstrates reading workbook properties from an XLSX or XLSB file.
+//!
+//! This example reads the core and extended document properties (such as
+//! creator, application, and company) from a workbook.
```
jmcnamara, line 8: Omit the phrase "core and extended" here and on any public docs. They won't mean anything to most end users.

examples/README.md
```diff
@@ -16,6 +16,8 @@ This directory contains some example of Calamine usage.
 - `read_hyperlinks.rs`: Reads the hyperlinks defined in an XLSX worksheet,
   either by sheet name or by sheet index.
 - `read_picture_data.rs`: Reads pictures and their metadata from an XLSX file.
+- `read_properties.rs`: Reads the workbook properties (core and extended)
```
jmcnamara, line 19: Omit the phrase "core and extended" here.

jmcnamara (changes requested):

jmcnamara:
@Alex-ley-scrub Thanks for the work here. Just one minor change required.

src/xlsb/mod.rs
```diff
@@ -438,6 +442,24 @@ impl<RS: Read + Seek> Xlsb<RS> {
         )
     }
 
+    /// Get workbook document properties.
+    ///
+    /// Missing fields are returned as `None`.
+    pub fn workbook_properties(&mut self) -> Result<&WorkbookProperties, XlsbError> {
```
jmcnamara, line 448: Thinking about this again, it is probably better to return `WorkbookProperties` via a `clone()` rather than the reference `&WorkbookProperties` which could keep the workbook owned and could prevent other method calls.

`WorkbookProperties` isn't one of the performance critical or heavyweight internal structures that needs to be returned as a ref. Side note, there are probably inconsistencies of this ref/value return usage in other places in the code but that is a separate issue.

jmcnamara (changes requested):

jmcnamara:
@Alex-ley-scrub Ping on this. There are some requested changes outstanding.

jmcnamara:
@Alex-ley-scrub This will need to be rebased due to the quick-xml changes introduced in #711

PR #691, opened 2026-07-25
Add a `shared_strings()` accessor on `Xlsx` and `Xlsb`

krickert:
This PR is made to speed up remote calls using a method Apache Arrow uses to shrink structures pinned by heavy serialization.

With the accessor below, a remote gRPC client can dictionary-encode calamine's string cells; the test results below show a transfer savings of 4.6x (12.7x with zstd on top) while streaming a 105.7 MB workbook over 10 GbE as fast as calamine parses it in-process.

The savings hold for any client language: a Python client transfers the same 62 MB instead of 789 MB for that workbook, and on loopback the gRPC stream into Python already runs at python-calamine speed (4.8 s vs 5.6 s).

So I made the same run against the NYC 311 data the calamine README benchmarks with: it cuts the transfer 6.4x. Charts and full captures are in [grpc-calamine]'s [bench].

The change is a public accessor for the resolved shared-strings table, `pub fn shared_strings(&self) -> &[String]` on `Xlsx` and `Xlsb`, with doc examples and tests checking that every `DataRef::SharedString` borrow resolves into the table. An index-preserving `DataRef::SharedStringIndex(u32)` cell variant was the other option considered; the accessor is the smaller API surface.

Calamine's reader already resolves each cell's sst index (`xlsx/cells_reader.rs:643` in 0.36.0) and discards it, so a streaming consumer today has to re-intern every resolved borrow that streams past in order to rebuild a table calamine is already holding in memory. The accessor seems like the most direct route, but let me know if there's a better one.

## Measurements

Dictionary mode is an opt-in flag in [grpc-calamine], built today on the interning workaround; the accessor would remove the table rebuild, about a second of CPU on a million-row workbook. Every mode feeds the same order-sensitive digest, so no mode wins by dropping data. On the NYC file (the README's CSV converted with LibreOffice into the same 186 MB xlsx) the full row stream is 662 MB plain, 301 MB with the dictionary, and 104 MB with zstd on top; the 105.7 MB workbook goes 789, 173, and 62 MB. Total CPU lands slightly below the plain stream because neither side copies string bodies per cell.

Bytes on the socket per mode, both datasets

[grpc-calamine]: https://github.com/ai-pipestream/grpc-calamine
[bench]: https://github.com/ai-pipestream/grpc-calamine/blob/main/bench/RESULTS.md

### DIFF
--- a/src/xlsb/mod.rs
+++ b/src/xlsb/mod.rs
@@ -419,30 +419,7 @@ impl<RS: Read + Seek> Xlsb<RS> {
     }
 
     /// Get the workbook's shared-strings table.
-    ///
-    /// XLSB files store every distinct cell string once, in a shared-strings
-    /// table, and cells reference entries by index. The table is resolved when
-    /// the workbook is opened, and each
-    /// [`DataRef::SharedString`](crate::DataRef::SharedString) borrows from
-    /// one of its entries. Exposing the table lets consumers that re-serialize
-    /// cells, such as a server streaming a worksheet over RPC, send each
-    /// string once and reference it by index instead of repeating string
-    /// bodies per cell.
-    ///
-    /// The table can be empty: workbooks that only use inline strings do not
-    /// have one.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// use calamine::{open_workbook, Error, Xlsb};
-    ///
-    /// fn main() -> Result<(), Error> {
-    ///     let workbook: Xlsb<_> = open_workbook("tests/date_1904.xlsb")?;
-    ///     println!("{} distinct strings", workbook.shared_strings().len());
-    ///     Ok(())
-    /// }
-    /// ```
+    #[doc(hidden)]
     pub fn shared_strings(&self) -> &[String] {
         &self.strings
     }
--- a/src/xlsb/mod.rs
+++ b/src/xlsb/mod.rs
@@ -418,6 +418,35 @@ impl<RS: Read + Seek> Xlsb<RS> {
         self.is_1904
     }
 
+    /// Get the workbook's shared-strings table.
+    ///
+    /// XLSB files store every distinct cell string once, in a shared-strings
+    /// table, and cells reference entries by index. The table is resolved when
+    /// the workbook is opened, and each
+    /// [`DataRef::SharedString`](crate::DataRef::SharedString) borrows from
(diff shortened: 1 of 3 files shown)


jmcnamara:
I'm not a fan of exposing the internal as a public function like this. It doesn't seem to be useful to an end user apart from this particular case.

Also, will the user need to parse the rich strings in the SST and handle any leading/trailing whitespace?

The `tests/test.rs` file has changed on main so you will need to resolve the conflicts.

@jqnatividad what do you think?

krickert:
On rich strings: no extra handling needed. The entries come from the same code path as cell values, so runs are already flattened and the text is identical to `Data::String`.

We are pushing on this because it means 10x less traffic for a serialized service. It's a trick Arrow does and commonplace on serialization methods today: the SST is the dictionary the format already provides, and without it every cell occurrence ships the full string body.

We marked it `#[doc(hidden)]` to make clear it's for internal serialization.

jqnatividad:
Thanks for the benchmarks @krickert , the dictionary win is real and the writeup is appreciated. But I'm with @jmcnamara on not taking this as-is.

The main thing: `DataRef::SharedString` gives you the resolved `&str`, not the index. Both readers resolve the sst index and throw it away. So having the table doesn't save you the str→index lookup per cell, which is the expensive part. Your own test shows the shape of it:

```rust
let table = excel.shared_strings().to_vec();
...
assert!(table.iter().any(|t| t == s), "{s:?} not in the table");
```

And that `.to_vec()` is forced - the cells reader borrows `&mut self` for the whole stream, so the returned slice can't be alive next to it. The accessor can't be used zero-copy in the case it exists for.

Second, `strings` isn't the file's SST, it's our normalized view of it: rich runs flattened, phonetic text dropped, and #671 just changed whitespace handling for `<t>` without `xml:space="preserve"`. Inline strings never land in it at all. You're right that the text matches `Data::String` today, but that's a moving target - anyone caching a dictionary across versions would drift on a patch bump, and we'd effectively be freezing normalization decisions we currently treat as fixable bugs. `#[doc(hidden)]` doesn't change that; it's still semver-public.

Worth noting you may not need us at all here: `DataRef<'a>` borrows into the table, so you can build a `HashMap<&'a str, u32>` as cells stream by, with no allocation and no re-interning, for the life of one reader.

krickert:
You are right - I didn't see that.  Thanks for the suggestion -

PR #695, opened 2026-07-27
xlsx: normalise reversed dimension refs

### LINKED ISSUE #692: Reversed `<dimension ref>` underflows unchecked u32 subtraction in get_dimension and Dimensions::len
A `<dimension>` whose end cell precedes its start underflows two unchecked `u32` subtractions. ECMA-376 does not require `ref` to be ordered, so this is reachable from a file a writer can legitimately emit, and from any untrusted input.

Two sites, same root, both still present on master:

- `src/xlsx/mod.rs:2789-2790` — `let rows = parts[1].0 - parts[0].0;` and the column line under it
- `src/lib.rs:181-183` — `Dimensions::len()` → `(self.end.0 - self.start.0 + 1) as u64 * ...`

### Repro

A minimal OPC package (one sheet, no shared strings, no styles) whose `xl/worksheets/sheet1.xml` is:

```xml
<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
<dimension ref="C5:A1"/>
<sheetData>
<row r="5"><c r="C5"><v>9</v></c></row>
</sheetData>
</worksheet>
```

```rust
let mut wb: Xlsx<_> = open_workbook(&path)?;
let name = wb.sheet_names()[0].clone();
let dims = wb.worksheet_cells_reader(&name)?.dimensions();
println!("declared start={:?} end={:?}", dims.start, dims.end);
println!("Dimensions::len() = {}", dims.len());
```

Debug build (calamine 0.36.0):

```
thread 'main' panicked at src/xlsx/mod.rs:2789:24:
attempt to subtract with overflow
```

Release build:

```
declared start=(4, 2) end=(0, 0)
Dimensions::len() = 18446744056529682435
```

### Why it is worth fixing rather than documenting

`worksheet_range` masks this in release, because `from_sparse` rebuilds the extent from the cells it actually saw. The damage surfaces through `dimensions()`, which is the natural pre-allocation hint for the incremental readers, and through `Dimensions::len()`, which is public API with no documented precondition. Reading the declared extent and calling `len()` on it is the obvious thing to do, and it silently yields 1.8e19.

This is the same shape as #150 and #174, which were fixed in the xls path. The xlsx path and the public `Dimensions` type still have it.

### Suggested fix

Normalise the corners in `get_dimension`, or reject the ref outright, or make the public computation total:

```rust
pub fn len(&self) -> u64 {
    if self.end.0 < self.start.0 || self.end.1 < self.start.1 {
        return 0;
    }
    u64::from(self.end.0 - self.start.0 + 1) * u64::from(self.end.1 - self.start.1 + 1)
}
```

Found while building the gRPC service in #688, which streams cells through `worksheet_cells_reader` and reports `dimensions()` to clients. Glad to send a PR if you have a preference for which of the three directions you want.

krickert:
Fixes #692.

A `<dimension>` whose end cell precedes its start, like `C5:A1`, underflowed the u32 subtractions in `get_dimension`: a debug build panics and a release build reports `Dimensions::len()` values around 1.8e19.

`Dimensions::new` now orders the corners so `start <= end` on both axes, and `get_dimension` goes through it, so a reversed `ref` parses the same as its ordered form. The `MAX_ROWS` / `MAX_COLUMNS` warnings compute from the normalised end corner. `Dimensions::len()` also widens to u64 before the `+ 1`, so a full axis no longer overflows.

A `Dimensions` built as a struct literal still bypasses the ordering; making the fields private is left to the wider API discussion from the review.

Tests: `test_reversed_dimension_is_normalised` (reversed on both axes and on each axis alone) and `test_dimensions_new_normalises_order`.

Touches the same `get_dimension` arm as #696; whichever lands second gets rebased.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,21 +169,18 @@ pub struct Dimensions {
 
 #[allow(clippy::len_without_is_empty)]
 impl Dimensions {
-    /// create dimensions info with start position and end position
-    ///
-    /// The corners may be given in either order; they are stored so that
-    /// `start <= end` on both axes.
+    /// Create a `Dimensions` instance with start and end cell positions.
     pub fn new(start: (u32, u32), end: (u32, u32)) -> Self {
         Self {
             start: (start.0.min(end.0), start.1.min(end.1)),
             end: (start.0.max(end.0), start.1.max(end.1)),
         }
     }
-    /// check if a position is in it
+    /// Check if a cell position is within the `Dimensions` range.
     pub fn contains(&self, row: u32, col: u32) -> bool {
         row >= self.start.0 && row <= self.end.0 && col >= self.start.1 && col <= self.end.1
     }
-    /// len
+    /// Get the `len()` of the `Dimensions` range, which is equivalent to the cell area.
     pub fn len(&self) -> u64 {
         // Widened before the `+ 1` so a full-width axis cannot overflow.
         (u64::from(self.end.0 - self.start.0) + 1) * (u64::from(self.end.1 - self.start.1) + 1)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -170,8 +170,14 @@ pub struct Dimensions {
 #[allow(clippy::len_without_is_empty)]
 impl Dimensions {
     /// create dimensions info with start position and end position
+    ///
+    /// The corners may be given in either order; they are stored so that
+    /// `start <= end` on both axes.
     pub fn new(start: (u32, u32), end: (u32, u32)) -> Self {
-        Self { start, end }
+        Self {
+            start: (start.0.min(end.0), start.1.min(end.1)),
+            end: (start.0.max(end.0), start.1.max(end.1)),
+        }
     }
     /// check if a position is in it
     pub fn contains(&self, row: u32, col: u32) -> bool {
@@ -179,7 +185,8 @@ impl Dimensions {
     }
     /// len
     pub fn len(&self) -> u64 {
-        (self.end.0 - self.start.0 + 1) as u64 * (self.end.1 - self.start.1 + 1) as u64
+        // Widened before the `+ 1` so a full-width axis cannot overflow.
+        (u64::from(self.end.0 - self.start.0) + 1) * (u64::from(self.end.1 - self.start.1) + 1)
     }
 }
 
--- a/src/xlsx/mod.rs
+++ b/src/xlsx/mod.rs
@@ -2789,18 +2789,21 @@ pub(crate) fn get_dimension(dimension: &[u8]) -> Result<Dimensions, XlsxError> {
             end: parts[0],
         }),
         2 => {
-            let rows = parts[1].0 - parts[0].0;
-            let columns = parts[1].1 - parts[0].1;
-            if rows > MAX_ROWS {
-                warn!("xlsx has more than maximum number of rows ({rows} > {MAX_ROWS})");
+            // The `ref` may be in reversed order like `C5:A1`.
+            let dim = Dimensions::new(parts[0], parts[1]);
+            if dim.end.0 > MAX_ROWS {
+                warn!(
+                    "xlsx has more than maximum number of rows ({} > {MAX_ROWS})",
+                    dim.end.0
+                );
             }
-            if columns > MAX_COLUMNS {
-                warn!("xlsx has more than maximum number of columns ({columns} > {MAX_COLUMNS})");
+            if dim.end.1 > MAX_COLUMNS {
+                warn!(
+                    "xlsx has more than maximum number of columns ({} > {MAX_COLUMNS})",
+                    dim.end.1
+                );
             }
-            Ok(Dimensions {
-                start: parts[0],
-                end: parts[1],
-            })
+            Ok(dim)
         }
         len => Err(XlsxError::DimensionCount(len)),
     }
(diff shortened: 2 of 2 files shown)


src/xlsx/mod.rs
```diff
@@ -2789,18 +2789,22 @@ pub(crate) fn get_dimension(dimension: &[u8]) -> Result<Dimensions, XlsxError> {
             end: parts[0],
         }),
         2 => {
-            let rows = parts[1].0 - parts[0].0;
-            let columns = parts[1].1 - parts[0].1;
+            // A `ref` is not required to be ordered, and writers do emit
+            // reversed ones such as `C5:A1`. Normalise the corners so the rest
+            // of the crate can rely on `start <= end`; subtracting them the
+            // other way round underflows, which panics in a debug build and
+            // wraps silently in a release one.
```
jmcnamara: This comment is overly verbose and seems AI written, here and elsewhere. Please review what is being submitted and make sure the comments are concise and to the point. In this case something like this.

```rust
// The `ref` may be in reversed order like `C5:A1`.
```

src/xlsx/mod.rs
```diff
@@ -2789,18 +2789,22 @@ pub(crate) fn get_dimension(dimension: &[u8]) -> Result<Dimensions, XlsxError> {
...
+            let start = (parts[0].0.min(parts[1].0), parts[0].1.min(parts[1].1));
+            let end = (parts[0].0.max(parts[1].0), parts[0].1.max(parts[1].1));
+            let rows = end.0 - start.0;
+            let columns = end.1 - start.1;
             if rows > MAX_ROWS {
```
jmcnamara: This should probably be `if end.0 > MAX_ROWS`. 

Applies to the next check too.

src/xlsx/mod.rs
```diff
@@ -4010,6 +4014,47 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_reversed_dimension_is_normalised() {
+        // ECMA-376 does not require `ref` to be ordered. A reversed one used to
+        // underflow the extent arithmetic: `C5:A1` panicked in a debug build,
+        // and in a release build produced `start: (4, 2), end: (0, 0)`, whose
+        // `len()` was then 18_446_744_056_529_682_435.
```
jmcnamara: Again, this is overly verbose and is talking in the past tense about panics that won't happen once the fix is in place. Just state what is being tested.

Also, is it a fact that "ECMA-376 does not require `ref` to be ordered"? Do you have a reference for that.

src/lib.rs
```diff
@@ -177,9 +177,20 @@ impl Dimensions {
...
     }
-    /// len
+    /// Number of cells covered by the area.
+    ///
+    /// Returns 0 for a degenerate area, i.e. one whose `end` precedes its
+    /// `start` on either axis. `Dimensions` is public and constructible
+    /// directly, and worksheets do carry reversed declarations, so this cannot
+    /// assume the corners are ordered: subtracting them the wrong way round
+    /// underflows, which panics in a debug build and wraps silently in a
+    /// release one.
```
jmcnamara: This is just hiding the real error. It shouldn't be possible to construct a "degenerate" Dimension. The constructor should either handle reordering or raise an error.

jmcnamara (changes requested):

jmcnamara:
The commit message is too verbose, here and in the other PRs that you submitted. Please write these yourself to describe concisely what has been changed or what is being fixed. You should be able to do that (for most changes) in 1-2 lines in the body.

krickert:
Bad assumption on my part since ECMA-376 isn't clear. The closest authoritative statement is [MS-OI29500 §2.1.1090](https://learn.microsoft.com/en-us/openspecs/office_standards/ms-oi29500/fe7f7a7d-5ec7-41f8-8631-3ac7884f0883), which says the standard doesn't clearly define what a valid cell range reference is, and that Office writes the smallest cell first. I've removed it. NBD since the normalization stands because it's a no-op on well-formed refs.

The constructor now normalizes the corners. The struct fields are still public, so a hand-built `Dimensions { start, end }` literal can bypass that. Should I make the fields private so that the invariant would be enforced?

Commit msg is now small. Working on the others.

src/xlsx/mod.rs
```diff
@@ -2789,18 +2789,21 @@ pub(crate) fn get_dimension(dimension: &[u8]) -> Result<Dimensions, XlsxError> {
...
         2 => {
-            let rows = parts[1].0 - parts[0].0;
-            let columns = parts[1].1 - parts[0].1;
-            if rows > MAX_ROWS {
-                warn!("xlsx has more than maximum number of rows ({rows} > {MAX_ROWS})");
+            // The `ref` may be in reversed order like `C5:A1`.
+            let dim = Dimensions::new(parts[0], parts[1]);
+            if dim.end.0 > MAX_ROWS {
+                warn!(
+                    "xlsx has more than maximum number of rows ({} > {MAX_ROWS})",
```
jmcnamara: While you are fixing this you might as well improve this error message. It is the file that exceeds the max row/col and not `xlsx`. Something like this would be better:

    "file has more than maximum supported number of rows ({} > {MAX_ROWS})"

src/lib.rs
```diff
@@ -170,16 +170,23 @@ pub struct Dimensions {
 #[allow(clippy::len_without_is_empty)]
 impl Dimensions {
     /// create dimensions info with start position and end position
+    ///
+    /// The corners may be given in either order; they are stored so that
+    /// `start <= end` on both axes.
```
jmcnamara: Remove this part of the comment. It isn't relevant to the end user.

Also, while you are improving this you could improve the meaning/punctuation of the comments:

    Create a `Dimensions` instance with start and end cell positions.
    Check if a cell positions in within the  `Dimensions` range.
    Get the `len()` of the `Dimensions` range, which is equivalent to the cell area.

jmcnamara (changes requested):

jmcnamara:
> Should I make the fields private so that the invariant would be enforced?

That would be an API change so it would need a wider discussion. However, in my opinion they should be private and replaced by accessors, and there shouldn't be any `len()` method.

Also, once you have addressed the required changes can you mark them as resolved so it is clear that the work is done.

src/lib.rs
krickert: I'll get to all the PRs later today. Thanks for looking into it

krickert:
Review comments addressed, let me know what you think.

PR #696, opened 2026-07-27
xlsx: error on cell references beyond the grid limits

### LINKED ISSUE #694: Cell reference parsing overflows u32 unchecked; MAX_ROWS/MAX_COLUMNS are warnings only
`get_row_and_optional_column` (`src/xlsx/mod.rs:2825-2860`) accumulates both components with unchecked `u32` arithmetic:

```rust
c @ b'A'..=b'Z' => col = col * 26 + (c - b'A') as u32 + 1,   // :2834
...
c @ b'0'..=b'9' => row = row * 10 + (c - b'0') as u32,        // :2850
```

Neither is bounded, and the `MAX_ROWS` / `MAX_COLUMNS` checks in `get_dimension` (`src/xlsx/mod.rs:2791-2796`) are `warn!` only, so an out-of-range value is logged and then returned to the caller anyway.

Three separate outcomes, all from ~2 KB packages with a single real cell. calamine 0.36.0.

### 1. In range for `u32`, far out of range for a spreadsheet

`<dimension ref="A1:ZZZZZZ1"/>`:

```
start=(0, 0) end=(0, 321272405)    (MAX_COLUMNS = 16384)
```

Same in debug and release. The `warn!` fires, then the value is handed back from `dimensions()`. That is roughly 19,600 times the grid width, and `dimensions()` is the natural pre-allocation hint for the incremental readers, so a caller that sizes a row buffer from it commits about 10 GB.

### 2. Column overflow

`<dimension ref="A1:ZZZZZZZ1"/>` (seven letters, 26^7 > `u32::MAX`):

```
debug:   thread 'main' panicked at src/xlsx/mod.rs:2834:38:
         attempt to multiply with overflow
release: start=(0, 0) end=(0, 4058115285)
```

### 3. Row overflow, and it is reachable from a cell reference

`<dimension ref="A1:A99999999999"/>` panics the same way at `:2850`. More importantly the same parser handles `<c r="...">`, so it is reachable while iterating cells rather than only when reading the declaration:

```xml
<dimension ref="A1:A5"/>
<sheetData>
<row r="99999999999"><c r="A99999999999"><v>2</v></c></row>
</sheetData>
```

```rust
let mut r = wb.worksheet_cells_reader(&name)?;
while let Some(c) = r.next_cell()? { println!("{:?}", c.get_position()); }
```

```
debug:   cell at (0, 0)
         thread 'main' panicked at src/xlsx/mod.rs:2850:38:
         attempt to multiply with overflow
release: cell at (0, 0)
         cell at (1215752190, 0)
```

In release the reader reports a cell at row 1,215,752,190, which is not a position that appears anywhere in the file.

### Why it compounds

A fabricated position like that is exactly what `Range::from_sparse` turns into an unbounded densification, so this feeds directly into #693: the two together mean a small file can produce a cell position no writer wrote, and then an allocation sized from it. Bounding the parse would close one of the two ways to reach that.

For what it is worth, `Reference::validate` (`src/xlsx/mod.rs:3122-3145`) already does exactly the right check with `XlsxError::ColumnNumberOverflow` and `RowNumberOverflow`, but it is only applied to formula references, not to `<dimension>` or `<c r>`.

### Suggested fix

Use checked arithmetic in the accumulation loops and return `XlsxError::ColumnNumberOverflow` / `RowNumberOverflow` on overflow or on exceeding the existing constants, reusing the variants `validate` already defines. That turns all three cases above into a normal `Result` the caller can handle, instead of a debug panic, a wrapped value, or a silently out-of-range one.

Found while building the gRPC service in #688. Related: #693 (unbounded densification, which this can feed) and #692 (unchecked subtraction on a reversed `ref`). Happy to send a PR.

krickert:
Fixes #694.

Both accumulators in `get_row_and_optional_column` were unchecked u32, so a reference like `ZZZZZZZ1` or `A99999999999` panicked in debug builds and wrapped in release builds, yielding cell positions that appear nowhere in the file.

As requested in review, the fix rejects invalid references instead of guarding the arithmetic: row and column accumulation is checked, and a reference past the grid (row > 1,048,576 or column > 16,384) returns the existing `XlsxError::RowNumberOverflow` / `ColumnNumberOverflow`. Dimension parsing propagates the same errors, replacing the warn-and-continue path in `get_dimension`. The column step moved into `push_column_letter` so the uppercase and lowercase arms stay identical.

`test_reference_beyond_grid_is_an_error` covers one past the grid on each axis, references that overflow u32 entirely, the same through `get_dimension`, and the bottom-right cell `XFD1048576`, which still parses.

Touches the same `get_dimension` arm as #695; whichever lands second gets rebased.

### DIFF
--- a/src/xlsx/mod.rs
+++ b/src/xlsx/mod.rs
@@ -2858,13 +2858,15 @@ fn get_row_and_optional_column(range: &[u8]) -> Result<(u32, Option<u32>), XlsxE
         i += 1;
     }
 
-    // Convert from 1-based to 0-based (col=0 means no column found)
+    // Reject references outside the maximum grid size.
     if row > MAX_ROWS {
         return Err(XlsxError::RowNumberOverflow);
     }
     if col > MAX_COLUMNS {
         return Err(XlsxError::ColumnNumberOverflow);
     }
+
+    // Convert from 1-based to 0-based (col=0 means no column found)
     let row = row
         .checked_sub(1)
         .ok_or(XlsxError::RangeWithoutRowComponent)?;
--- a/src/xlsx/mod.rs
+++ b/src/xlsx/mod.rs
@@ -2788,20 +2788,10 @@ pub(crate) fn get_dimension(dimension: &[u8]) -> Result<Dimensions, XlsxError> {
             start: parts[0],
             end: parts[0],
         }),
-        2 => {
-            let rows = parts[1].0 - parts[0].0;
-            let columns = parts[1].1 - parts[0].1;
-            if rows > MAX_ROWS {
-                warn!("xlsx has more than maximum number of rows ({rows} > {MAX_ROWS})");
-            }
-            if columns > MAX_COLUMNS {
-                warn!("xlsx has more than maximum number of columns ({columns} > {MAX_COLUMNS})");
-            }
-            Ok(Dimensions {
-                start: parts[0],
-                end: parts[1],
-            })
-        }
+        2 => Ok(Dimensions {
+            start: parts[0],
+            end: parts[1],
+        }),
         len => Err(XlsxError::DimensionCount(len)),
     }
 }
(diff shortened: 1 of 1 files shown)


jmcnamara:
This PR is fixing the wrong thing. Rather than handling U32 overflow it should be handling invalid row/column dimensions and raising an error.

krickert:
Done.
Reworked the fix in 15fad0f:  `get_row_and_optional_column` now validates against the grid and returns `XlsxError::RowNumberOverflow` (row > 1,048,576) or `XlsxError::ColumnNumberOverflow` (column > 16,384) instead of only guarding the u32 bounds.

The lenient warn! path in `get_dimension` is gone so dimension parsing propagates the error. The checked arithmetic is retained for the beyond-u32 case. 

`test_reference_beyond_grid_is_an_error` covers one-past-grid on both axes, `u32` overflow, and the in-bounds corner `XFD1048576`.

krickert:
Reworked to reject invalid references as requested, let me know what you think.

Issue #684, opened 2026-07-19
read_shared_strings reserves Vec<String> capacity from unbounded attacker-controlled uniqueCount


devsamhan:
## Summary

`Xlsx::read_shared_strings` (`src/xlsx/mod.rs`) reserves `Vec<String>` capacity directly from the attacker-controlled `uniqueCount` attribute in `sharedStrings.xml`, with no upper bound.

Affected version observed locally: `calamine 0.36.0`. I have not verified whether this is already fixed on `main`; this report is based on local source inspection of `calamine-0.36.0` only (no CI/test-suite check was performed against a newer revision).

## Affected code path

In `src/xlsx/mod.rs`, `read_shared_strings` parses the `<sst uniqueCount="N">` root attribute and calls `self.strings.reserve(n)`:

```rust
if let Some(count) = e.raw_attr(b"uniqueCount")? {
    if let Ok(n) = atoi_simd::parse::<usize, true, false>(count) {
        self.strings.reserve(n);
    }
}
```

This runs unconditionally inside `Xlsx::new()`'s constructor, before the caller regains control — there is no point between calling `open_workbook`/`open_workbook_auto` and getting a `Result` back where the caller's own code executes.

Contrast with `worksheet_range_ref`'s structurally identical `<dimension>`-driven reservation, which the same file already guards:

```rust
let len = cell_reader.dimensions().len();
let mut cells = Vec::new();
if len < 100_000 {
    cells.reserve(len as usize);
}
```

`read_shared_strings`'s `uniqueCount`-driven reservation has no equivalent guard.

## Impact

A ~1 KB, otherwise perfectly ordinary `.xlsx` (normal ZIP central directory, no unusual compression ratio, no oversized declared/actual size anywhere) can declare `uniqueCount="4000000000"` in `sharedStrings.xml` with zero `<si>` children, driving a single `Vec::reserve` request for ~96 GB (4,000,000,000 × 24 bytes/`String` header on a 64-bit target) before the caller ever regains control.

Because this uses the non-fallible `.reserve()` (not `try_reserve`), an allocation failure invokes `handle_alloc_error`, whose default behavior is to abort the process — not a catchable `Result`. A calling application's ordinary `Result`-based error handling around `open_workbook`/`open_workbook_auto` cannot intercept this; the effect is a crash of the entire host process.

## Minimal repro shape

[...]

PR #697, opened 2026-07-27
crate: return an error from Range::from_sparse instead of aborting

### LINKED ISSUE #693: Range::from_sparse densifies without bound: a 2 KB well-formed xlsx aborts the process
`Range::from_sparse` densifies to `rows * cols` with no bound (`src/lib.rs:958-961`):

```rust
let len = cols.saturating_mul(rows);
let mut v = vec![T::default(); len];
```

`saturating_mul` prevents the arithmetic overflow, but the allocation still runs at whatever size the corners imply. Two cells far apart is enough, and the failure is an **abort**, not a panic.

### Repro

A 2,341 byte package. Both cells are legal: `XFD1048576` is the last addressable cell in the grid, so nothing here is malformed.

`xl/worksheets/sheet1.xml`:

```xml
<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
<dimension ref="A1:XFD1048576"/>
<sheetData>
<row r="1048576"><c r="XFD1048576"><v>2</v></c></row>
</sheetData>
</worksheet>
```

```rust
let mut wb: Xlsx<_> = open_workbook(&path)?;
let name = wb.sheet_names()[0].clone();
let r = wb.worksheet_range(&name)?;   // never returns
```

Run under a 4 GB cap so it fails fast rather than thrashing the machine (calamine 0.36.0, release):

```
$ ( ulimit -v 4000000; ./repro corners.xlsx; echo "exit=$?" )
memory allocation of 549755813888 bytes failed
Aborted (core dumped)
exit=134
```

1,048,576 x 16,384 = 17,179,869,184 cells, and `size_of::<Data>()` is 32, so 549.7 GB. On a machine with enough overcommit it will try to clone-fill every one of those pages instead.

### The part that makes this different from a large-allocation bug

Allocation failure calls `handle_alloc_error`, which aborts. It does not unwind. So a library consumer cannot defend against it on stable Rust:

- `catch_unwind` does not catch it
- running the parse on its own thread does not help; the whole process goes
- exit 134 is SIGABRT, visible above

The only stable-Rust defence is to not call the function. That is workable for cell values, because `worksheet_cells_reader` avoids `from_sparse` entirely. It is not workable for formulas: `worksheet_formula` goes through `from_sparse` at `src/xlsx/mod.rs:2617` and has no incremental equivalent, so a service that reads formulas from uploaded files has no way to protect itself.

This is the same failure mode as #444 ("memory allocation of 2199023255552 bytes failed"), #550 and #433, which were addressed in other paths.

### Prior art

#463 raised the `from_sparse` OOM and proposed a fix. Reading the thread, it was not turned down: you replied positively and asked for a clarifying comment, sftse added it and said it was ready, and the PR then went quiet and was self-closed nine months later. Worth noting that sftse's stated workaround was `set_alloc_error_hook`, which is unstable, which is consistent with there being no way to handle this on stable.

I did not want to simply reopen that PR without checking whether you want the same approach.

### Possible directions

1. Return `Err` when `len` crosses a threshold, so the caller sees a normal error instead of losing the process.
2. Allocate fallibly (`try_reserve` plus `resize`) and map failure to an error.
3. Bound by density rather than absolute size, so a genuinely large dense sheet still loads but a two-cell sheet spanning the grid does not.

Any of the three would fix it from a consumer's point of view. Happy to send a PR for whichever you prefer.

Found while building the gRPC service in #688. Related: #684, which is the same class of attacker-influenced allocation on a different code path.

krickert:
Fixes #693.

A `Range` is dense, so the extent implied by the outermost cell positions decides the allocation, not the number of cells. A sheet holding only `A1` and `XFD1048576` implies 17,179,869,184 cells, and `vec![T::default(); len]` aborts the process when that allocation fails, which a caller cannot catch.

As requested in review, `from_sparse` now returns `Result<Range<T>, RangeError>`. The buffer is reserved with `try_reserve_exact`, and the error carries the implied width and height. The xls, xlsb and xlsx readers propagate it through a new `Range` variant on their error enums, so `worksheet_range` returns `Err` instead of killing the process. The extent arithmetic is widened to u64 before the `+ 1`, which was a second overflow route on 32-bit targets.

This changes the signature of a public function; the doc examples are updated to match.

Tests: an oversized two-corner extent returns the error with its dimensions, `tests/issue_693.xlsx` exercises the same through the xlsx reader, and a sparse-but-large extent still loads, including `issue_174.xlsx`, which declares the full grid but densifies to 2x11.


jmcnamara:
This one also came up in the initial review for #684. I may fix it as part of that. If not I will come back to this.

Possibly this should be handled like issue #594 and fix #596.

src/lib.rs
```diff
@@ -917,6 +917,15 @@ impl<T: CellType> Range<T> {
     ///
     /// - `cells`: A vector of [`Cell`] elements.
     ///
+    /// # Panics
```
jmcnamara: As a general (Rust) rule we shouldn't be fixing issues by panicking, even if they are documented. The correct (but harder) fix it to introduce an `Error` to the return type.

src/lib.rs
```diff
@@ -955,10 +964,31 @@ impl<T: CellType> Range<T> {
...
-        let rows = (row_end - row_start + 1) as usize;
-        let len = cols.saturating_mul(rows);
-        let mut v = vec![T::default(); len];
+        // Widened before the `+ 1`: an extent spanning a full `u32` axis would
+        // overflow otherwise, and `usize` is 32 bits on wasm32.
+        let width = u64::from(col_end - col_start) + 1;
+        let height = u64::from(row_end - row_start) + 1;
+        let size = width.saturating_mul(height);
+
+        // The extent comes from the cell positions in the file, so two cells
```
jmcnamara: Similar to comments on the other PRs. This looks machine generated and overly verbose. Please review and add human judgement.

jmcnamara (changes requested):

krickert:
from_sparse now returns a Result as requested, let me know what you think.

PR #699, opened 2026-07-28
fix(xls/xlsb): correct column letters in reconstructed formulas (columns >= AA)

binggao1230:
### Problem

`utils::push_column` converts a 0-based column index to Excel column letters, but it used plain base-26 and dropped the most-significant digit, so **every index >= 26 was wrong**:

| index | got | Excel |
|-------|-----|-------|
| 26 | `A` | `AA` |
| 52 | `A` | `BA` |
| 701 | `AZ` | `ZZ` |
| 702 | `BA` | `AAA` |

775 of the first 801 indices disagree with Excel; only `A`..`Z` (0..25) were correct.

`push_column` is the renderer behind all ~13 column references emitted by the `.xls` and `.xlsb` formula readers, so `worksheet_formula` returned **silently wrong columns** for any binary-format formula referencing column `AA` or beyond, e.g. `=SUM(AA1:AA10)` came back as `=SUM(A1:A10)` — a different column, no error. (xlsx/ODS are unaffected; they carry formulas as strings.)

Excel columns are bijective base-26, and the crate already implements this correctly in `xlsx::column_number_to_name` (with a passing test). `push_column` just never matched its sibling. The fix mirrors that function; a new test asserts the two are byte-identical across the whole valid column range.

### Second bug in the same path

While adding an integration fixture I found the BIFF8 `PtgArea` branch passed the raw column word — including the `fColRel`/`fRwRel` flag bits — straight to `push_column` and hardcoded `$`, so `SUM(AA1:AA3)` rendered as `$BTSM$1:$BTSM$3`. It now masks the 14-bit column and reads the relative flags, mirroring the adjacent `PtgRef` branch that already does this.

The 3D-reference branches (`PtgRef3d`/`PtgArea3d`) have the same unmasked-column pattern and a pre-existing `// TODO: check with relative columns`; I left those for a separate change to keep this focused.

### Tests

- `push_column_is_bijective_base26` — boundary cases (`0`->`A`, `25`->`Z`, `26`->`AA`, `701`->`ZZ`, `702`->`AAA`, `16383`->`XFD`) plus a differential check against `xlsx::column_number_to_name` over the full column range.
- `xls_formula_columns_beyond_z` — a new `.xls` fixture whose formulas reference columns >= 26 (existing formula fixtures only use single-letter columns, so this path was untested). Renders `SUM(AA1:AA3)` and `AA1+AB1`.

Full suite passes.

This PR was written with AI assistance under my direction; I have reviewed it and am accountable for it.


src/utils.rs
```diff
@@ -65,20 +65,18 @@ pub fn read_f64(s: &[u8]) -> f64 {
...
-            let c = col % 26;
-            rev.push((b'A' + c as u8) as char);
-            col -= c;
-            col /= 26;
-        }
-        buf.extend(rev.chars().rev());
+/// Push a 0-based column index as its Excel letters (0 -> "A", 26 -> "AA").
+///
+/// Excel columns are bijective base-26, so every digit contributes; the plain
+/// base-26 the previous code used dropped the most-significant digit.
```
jmcnamara: Omit this part of the comment. It is referring to the "previous code" which doesn't make sense. Also, while "bijective" is correct it is an unusual word in this context. The full comment would be better as something like:

```rust
// Convert a 0-based column index to an Excel column name (0 -> "A", 26 -> "AA").
```

src/utils.rs
```diff
@@ -65,20 +65,18 @@ pub fn read_f64(s: &[u8]) -> f64 {
...
+pub fn push_column(col: u32, buf: &mut String) {
```
jmcnamara, line 69: As a general comment this function would be better to be replaced by `column_number_to_name()` or vice-versa.

src/utils.rs
```diff
@@ -65,20 +65,18 @@ pub fn read_f64(s: &[u8]) -> f64 {
...
+    let mut num = col + 1;
+    let mut rev = String::new();
```
jmcnamara: This change creates a new buffer for every function call, even when it isn't required.

src/utils.rs
```diff
@@ -1165,6 +1163,35 @@ pub const FTAB_ARGC: [u8; FTAB_LEN] = [
 mod tests {
     use super::*;
 
+    // push_column must agree with the crate's own tested bijective converter
+    // (xlsx::column_number_to_name) across the whole valid column range.
```
jmcnamara: Omit or improve this comment.

src/utils.rs
```diff
@@ -1165,6 +1163,35 @@ pub const FTAB_ARGC: [u8; FTAB_LEN] = [
...
+    #[test]
+    fn push_column_is_bijective_base26() {
```
jmcnamara: Just call the function `test_push_column()`

src/utils.rs
```diff
@@ -1165,6 +1163,35 @@ pub const FTAB_ARGC: [u8; FTAB_LEN] = [
...
+        check(16383, "XFD");
+
+        // Differential oracle: byte-identical to xlsx::column_number_to_name.
+        for col in 0..crate::xlsx::MAX_COLUMNS {
+            let mut got = String::new();
+            push_column(col, &mut got);
+            let mut want = Vec::new();
+            crate::xlsx::column_number_to_name(col, &mut want).unwrap();
+            assert_eq!(got.as_bytes(), &want[..], "diverged at col {col}");
+        }
```
jmcnamara: Omit the comparison test with `column_number_to_name()`. As I said it another comment it would probably be best to just use that function anyway.

src/xls.rs
```diff
@@ -1753,11 +1753,27 @@ fn parse_formula(
                     write!(&mut formula, "${row_last}").unwrap();
                     rgce = &rgce[6..];
                 } else {
-                    formula.push('$');
-                    push_column(read_u16(&rgce[4..6]) as u32, &mut formula);
-                    write!(&mut formula, "${}:$", read_u16(&rgce[0..2]) as u32 + 1).unwrap();
-                    push_column(read_u16(&rgce[6..8]) as u32, &mut formula);
-                    write!(&mut formula, "${}", read_u16(&rgce[2..4]) as u32 + 1).unwrap();
+                    // columnFirst/columnLast are ColRelU: 14-bit column + fColRel/fRwRel flags.
```
jmcnamara, line 1776: As a general comment this should be in a separate commit at least, and probably better as a separate bug report/PR.

tests/test.rs
```diff
@@ -3616,3 +3616,15 @@ fn xls_empty_string() {
     let range = wb.worksheet_range("Sheet1").unwrap();
     assert_eq!(range.get_value((0, 0)), Some(&String("".to_string())));
 }
+
+#[test]
+fn xls_formula_columns_beyond_z() {
+    // Formula column references at/after column AA (index 26) exercise the
+    // bijective base-26 rendering and the BIFF8 PtgArea column masking.
```
jmcnamara: Omit the word bijective.

.typos.toml
```diff
@@ -1,5 +1,6 @@
 [default.extend-words]
 typ = "typ"
+BA = "BA"
```
jmcnamara: Just omit `BA` from the test case or use another column name that doesn't trigger the spell checker.

jmcnamara (changes requested):

jmcnamara:
As a general comment for future submissions, fixes for an issue should start with a bug report. This change also contains two separate bugs/PRs.

binggao1230:
All applied: push_column and column_number_to_name now share one allocation-free digit-fill core, comments trimmed, test renamed to test_push_column (BA case and typos entry dropped), and the branch is split into two commits — fix(utils) and fix(xls). Kept them in one PR since the xls fix builds on the column rendering; happy to split into a separate PR if you prefer.

src/utils.rs
binggao1230: Done — column_number_to_name now delegates to the shared digit-fill core in utils, so there is one implementation.

src/utils.rs
binggao1230: Fixed — the per-call String is gone; digits are written into a stack array.

jmcnamara:
Thank you. Could you also mark the comments as resolved.

binggao1230:
All review threads are now marked resolved. Thanks!

jmcnamara:
Merged. Thanks.

PR #701, opened 2026-07-30
fix(xls): skip nested substreams so embedded charts do not leak cells

### LINKED ISSUE #700: Bug: embedded cross-sheet chart in .xls leaks its cached series into the worksheet as cells at A1
### Current behavior

When a legacy `.xls` (BIFF8) worksheet contains an **embedded chart whose series live on a different sheet**, the chart's cached series values are returned as *cells of the hosting worksheet*, starting at `A1`. They fill empty cells and **overwrite populated ones**, so real content is lost with no error.

An embedded chart is a nested substream inside the worksheet substream. The worksheet record loop in `src/xls.rs` has no arm for `BOF` (`0x0809`), so once the chart's `BOF` is passed, the chart's records keep being matched by the cell arms (`0x0203` Number, `0x027E` Rk, …) and pushed into `cells`.

On a real-world file this destroyed exactly the identifying header — a document number, a title line and three row labels all became floats, while the rest of the sheet read correctly. `xlrd` reads the same file correctly.

Confirmed with two controls:

- **Deleting the charts** in Excel and saving → reads correctly.
- **Re-saving with the charts kept** → still wrong. So it is the charts, not "Excel rewrote the file".
- A **same-sheet** chart does *not* trigger it. Only cross-sheet series do, which is why a naive "add a chart" test misses this.

Reproduced identically via `python-calamine` 0.6.2 (calamine 0.33.0) and 0.8.2 (calamine 0.36.0), and via `polars.read_excel`, so it is not specific to a binding or version.

### Expected behavior

Records belonging to a nested substream (embedded chart, macro sheet, …) are skipped rather than parsed as cells, so `worksheet_range` returns only the worksheet's own cells and populated cells are never overwritten.

### Sample code to reproduce

```rust
use calamine::{open_workbook, Data, Reader, Xls};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // "Report"!A1 contains the string "HEADER TEXT A1".
    // The sheet hosts a scatter chart whose series are Data!A1:B8.
    let mut wb: Xls<_> = open_workbook("x_cross_sheet.xls")?;
    let range = wb.worksheet_range("Report")?;

    // Expected: Some(&Data::String("HEADER TEXT A1"))
    // Actual:   Some(&Data::Float(1.0))   <- first cached chart series value
    println!("{:?}", range.get((0, 0)));
    assert_eq!(range.get((0, 0)), Some(&Data::String("HEADER TEXT A1".to_string())));
    Ok(())
}
```

### Test file

I can attach one — it is synthetic and contains no real data. To generate an equivalent file (Windows + Excel, via `xlwings`):

```python
import xlwings as xw

app = xw.App(visible=False, add_book=False)
wb = app.books.add()
ws = wb.sheets[0]; ws.name = 'Report'
ws.range('A1').value = 'HEADER TEXT A1'
ws.range('A6').value = 'LABEL A6'; ws.range('C6').value = 'VALUE C6'

data = wb.sheets.add('Data', after=ws)                 # series on ANOTHER sheet
data.range('A1').value = [[i, i * 1.5 + 0.25] for i in range(1, 9)]

ch = ws.charts.add(left=300, top=20, width=240, height=160)
ch.set_source_data(data.range('A1:B8'))                # cross-sheet reference
ch.chart_type = 'xy_scatter_lines'

wb.api.SaveAs(r'C:\tmp\x_cross_sheet.xls', FileFormat=56)   # 56 = xlExcel8 (BIFF8)
wb.close(); app.quit()
```

Point the chart at `ws` instead of `data` and the bug disappears — that switch is the whole trigger.

### Environment

- calamine 0.33.0 and 0.36.0
- rustc 1.97.1, x86_64-pc-windows-msvc
- also reproduced through `python-calamine` 0.6.2 / 0.8.2 and `polars.read_excel`
- source files written by Excel (BIFF8 `.xls`)

### Suggested fix

Track substream nesting in the worksheet record loop and skip nested records. The first `BOF` in the stream is the worksheet's own, since `BoundSheet8`'s position points at it. Against `v0.33.0`:

```diff
--- a/src/xls.rs
+++ b/src/xls.rs
@@ -441,9 +441,32 @@ impl<RS: Read + Seek> Xls<RS> {
             let mut merge_cells = Vec::new();
+            let mut seen_sheet_bof = false;
+            let mut substream_depth = 0usize;
             for record in records {
                 let r = record?;
+                if substream_depth > 0 {
+                    match r.typ {
+                        0x0809 => substream_depth += 1,
+                        0x000A => substream_depth -= 1,
+                        _ => (),
+                    }
+                    continue;
+                }
                 match r.typ {
+                    // BOF of a nested substream (embedded chart, macro sheet, ...)
+                    0x0809 => {
+                        if seen_sheet_bof {
+                            substream_depth += 1;
+                        } else {
+                            seen_sheet_bof = true;
+                        }
+                    }
                     // 512: Dimensions
                     0x0200 => {
```

With that applied, the affected real-world file matches `xlrd` on every cell, a same-sheet-chart file is unchanged, and other `.xls` files I tested are unaffected. Happy to open a PR with a test fixture if useful.

Related: dimastbk/python-calamine#161 reports the same symptom ("first two columns wrong in old xls files", values "that look like they come from elsewhere in the file"). It has been open since 2025-10 because the reporter's files are confidential and re-saving them in Excel makes the bug vanish — the cross-sheet-chart condition above explains why, and makes a shareable repro possible.

Nitjsefnie:
## Summary

The `.xls` worksheet record loop had no arm for `BOF`, so an embedded chart's cached series records were parsed as cells of the hosting sheet starting at A1 — filling empty cells and overwriting populated ones with no error. This tracks substream nesting and skips nested records.

## Related Issues and Pull Requests

```
Fixes #700
Relates to dimastbk/python-calamine#161
```

## Changes

- `src/xls.rs`: track substream nesting in the worksheet record loop. Records inside a nested substream are skipped instead of matching the cell arms (`0x0203` Number, `0x027E` Rk, …). The first `BOF` in the stream is the worksheet's own — `BoundSheet8`'s position points at it — hence the `seen_sheet_bof` flag rather than an unconditional depth increment.
- `tests/test.rs`: `xls_embedded_cross_sheet_chart_does_not_leak_cells` — asserts `A1`, `A6` and `C6` hold their real strings.
- `tests/xls_cross_sheet_chart.xls`: new fixture. Synthetic, no real data — `"Report"!A1` holds `HEADER TEXT A1` and the chart on that sheet plots `Data!A1:B8`.

Behaviour is unchanged for a chart whose series live on the *same* sheet as the chart; only cross-sheet series triggered the leak, which is why the obvious "add a chart" test passes without this fix.

## Testing

- `cargo test` with the patch applied: **158 passed, 0 failed, 1 ignored** (the pre-existing ignore) in the `test` target, plus **65 passed, 0 failed** in the second target.
- The new test fails without the patch (`A1` reads `Float(1.0)`) and passes with it.
- Verified against a real-world `.xls` that triggered this in production: with the patch its `Protokol` sheet matches `xlrd` on every cell except two date cells, where calamine returns `datetime` and `xlrd` returns raw serials — an intended difference, not a regression.
- Two controls confirming the cause rather than a correlate: deleting the charts in Excel and saving makes the unpatched reader correct, while re-saving with the charts *kept* leaves it wrong. So it is the charts, not Excel rewriting the file.
- Reproduced before the fix on 0.33.0 and 0.36.x, and through `python-calamine` 0.6.2 / 0.8.2 and `polars.read_excel`.

## Follow-ups / Known Limitations

- Only the BIFF (`.xls`) path is addressed. I did not test whether an equivalent leak exists for `.xlsx`, where charts are stored differently — if it does, it needs its own fix.
- The fixture was produced by Excel via COM, so regenerating it needs Windows + Excel. The recipe is in #700 if a differently-shaped fixture is preferred.
- The nesting counter is deliberately generic (it skips any nested substream, not only charts). If some nested substream type is expected to contribute cells, this would suppress it — I found no such case in the test corpus.

## Footer

Generated by Claude Opus 5 (1M context) (diagnosis, implementation, testing)


jmcnamara:
Mainly looks okay to me. The comments are too verbose and have a written by AI feel but apart from that it looks okay.

@sftse any concerns?

Nitjsefnie:
Trimmed. The block in `src/xls.rs` is two lines now and the test comment one sentence, matching the terse style already in those files — the diff is comments only, no code touched.

I kept one line, "The first BOF belongs to the worksheet itself", because it's the non-obvious bit that explains why `seen_sheet_bof` exists at all.

sftse:
@jmcnamara
I can't review within the next 10 days. I know you've been taking the brunt of AI-fueled PRs, I'd like to help out but would need to be at a slower pace.

jmcnamara:
@sftse No worries. This one looks okay to me, after the changes.

Nitjsefnie:
Don't worry, I won't be sending any more AI PRs to this repo, I just needed this fixed as it is a dependency in my app, so I thought why not send it as PR as well.

jmcnamara:
Merged. Thanks.

Issue #612, opened 2026-02-28
Panic in xls.rs when BIFF5 (.xls) file has named range referencing sheet index >= 9


Tim-Kracht:
## Summary

Calamine panics with `range end index 2 out of range for slice of length 1` at `xls.rs:388` when opening a BIFF5 (Excel 5.0/95) `.xls` file that contains a named range referencing a worksheet at sheet index 9 or higher.

## Steps to Reproduce

1. Open Excel and create a new workbook
2. Add sheets until you have at least 10 (Sheet1 through Sheet10)
3. Create a named range (Formulas → Name Manager → New) pointing to a cell on Sheet10 or later (any sheet at index >= 9)
4. **Save As** → **"Microsoft Excel 5.0/95 Workbook (.xls)"** format
5. Open the file with calamine

A minimal reproducer file is attached: `calamine-panic-empty-sheet.xls` — 11 empty sheets, one named range (`test`) pointing to a cell on Sheet10 (index 9).

## Rust reproducer

```rust
use calamine::{open_workbook, Reader, Xls};

fn main() {
    let mut workbook: Xls<_> =
        open_workbook("calamine-panic-empty-sheet.xls").expect("could not open workbook");

    println!("Sheet names: {:?}", workbook.sheet_names());
}
```

```
thread 'main' panicked at calamine-0.33.0/src/xls.rs:388:54:
range end index 2 out of range for slice of length 1
```

## Backtrace

```
stack backtrace:
   7: calamine::xls::Xls<RS>::parse_workbook::{{closure}}
             at calamine-0.32.0/src/xls.rs:388:54
  16: calamine::xls::Xls<RS>::parse_workbook
             at calamine-0.32.0/src/xls.rs:387:30
  17: calamine::xls::Xls<RS>::new_with_options
             at calamine-0.32.0/src/xls.rs:214:13
  18: <calamine::xls::Xls<RS> as calamine::Reader<RS>>::new
             at calamine-0.32.0/src/xls.rs:239:9
  19: calamine::open_workbook
             at calamine-0.32.0/src/lib.rs:409:5
  20: calamine_bug::main
             at ./src/main.rs:5:9
```

## Analysis

The named range formula in BIFF5 format uses a `ptgArea3d` token (`0x3a`) with sheet indices stored in the formula bytes. Through testing we determined:

- Named ranges referencing sheet indices **0–8** parse successfully
- Named ranges referencing sheet index **9 or higher** cause the panic
[...]

Issue #270, opened 2022-09-22
range end index 2 out of range for slice of length 0


1991766703:
thread 'main' panicked at 'range end index 2 out of range for slice of length 0', library\core\src\slice\index.rs:73:5
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\panicking.rs:584   
   1: core::panicking::panic_fmt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\panicking.rs:142  
   2: core::slice::index::slice_end_index_len_fail_rt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\slice\index.rs:73 
   3: core::ops::function::FnOnce::call_once<never$ (*)(usize,usize),tuple$<usize,usize> >
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\ops\function.rs:248
   4: core::intrinsics::const_eval_select<tuple$<usize,usize>,never$ (*)(usize,usize),never$ (*)(usize,usize),never$>
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\intrinsics.rs:2376   5: core::slice::index::slice_end_index_len_fail
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\slice\index.rs:67 
   6: core::slice::index::impl$3::index<u8>
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f\library\core\src\slice\index.rs:304   7: core::slice::index::impl$4::index<u8>
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f\library\core\src\slice\index.rs:351   8: core::slice::index::impl$0::index<u8,core::ops::range::RangeTo<usize> >
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f\library\core\src\slice\index.rs:18 
   9: calamine::utils::read_u16
             at C:\Users\Administrator\.cargo\registry\src\rsproxy.cn-8f6827c7555bfaf8\calamine-0.18.0\src\utils.rs:34
  10: calamine::xls::read_rich_extended_string
8.0\src\xls.rs:579
  11: calamine::xls::parse_sst
[...]

PR #707, opened 2026-08-10
xls: fix panic on a PtgExp token with a truncated operand

Butch78:
## Current behavior

`parse_formula` advances past a `PtgExp` token by slicing four bytes off `rgce` without checking that four bytes remain:

```rust
// src/xls.rs:1515
0x01 => {
    // PtgExp: array/shared formula, ignore
    debug!("ignoring PtgExp array/shared formula");
    stack.push(formula.len());
    rgce = &rgce[4..];        // ← no length check
}
```

A workbook whose `PtgExp` operand is truncated panics:

```
thread 'main' panicked at src/xls.rs:1515:29:
range start index 4 out of range for slice of length 2
```

Because formulas are parsed while the workbook is read, this aborts **`open_workbook` itself** — the file cannot be opened at all, let alone have its formulas inspected. Any caller handing calamine a file it did not produce (an upload, a corpus, a user attachment) is one malformed workbook away from losing the thread.

## Expected behavior

A truncated token is malformed *input*, not a bug in the caller, so it should surface as an `Err` like the reader's other length violations — and the rest of the workbook should still load.

## Sample code to reproduce

```rust
use calamine::{open_workbook, Reader, Xls};

fn main() {
    // panics before this returns
    let mut wb: Xls<_> = open_workbook("ptgexp-truncated-operand.xls").unwrap();
    println!("{:?}", wb.sheet_names());
}
```

## The fix

Report the short token as `XlsError::Len`, matching how the reader already handles other length violations (e.g. the `Formula` record check at `src/xls.rs:640`):

```rust
if rgce.len() < 4 {
    return Err(XlsError::Len {
        expected: 4,
        found: rgce.len(),
        typ: "PtgExp",
    });
}
rgce = &rgce[4..];
```

With the fix the attached workbook opens and yields real data — a 23×10 range and 22 formula cells — so this recovers the file rather than merely avoiding the crash. That is what the test asserts.

## Test file

`tests/ptgexp-truncated-operand.xls`, from the public Enron/EDRM corpus via SheetJS. It is the only file in a 2000-workbook sample that triggers this, which is presumably why it has not been reported before.

Test added to `tests/test.rs` in the existing style; it panics without the fix and passes with it.

## Environment

```
- calamine version: 0.36.1 (also reproduced on master @ 0a24c2a)
- Cargo.toml dependency line for calamine: calamine = "0.36"
- rustc version: nightly
- OS: Linux
```

## Checks

`cargo fmt -- --check`, `cargo clippy --all-targets --all-features -- -Dwarnings`, `cargo test --all-features`, and `cargo check --target wasm32-unknown-unknown --all-features` all pass locally.

## A note on the wider pattern

Every arm of `parse_formula` advances `rgce` with an unchecked slice, so this class of panic is not exhausted — #270, #447, #600, #612 and #643 were each the same shape in a different token. I have kept this PR to the one token I can demonstrate with a fixture rather than sweeping the function, since a broad change to a parser I do not maintain is harder to review. Happy to follow up with a systematic guard if you would find that useful.


src/xls.rs
```diff
@@ -1512,6 +1512,17 @@ fn parse_formula(
                 // PtgExp: array/shared formula, ignore
                 debug!("ignoring PtgExp array/shared formula");
                 stack.push(formula.len());
+                // PtgExp carries a 4-byte operand (the row/column of the
+                // array or shared formula it points at). A truncated token
+                // leaves fewer bytes than that, so slicing unconditionally
+                // panics instead of reporting the malformed input.
```
jmcnamara: This comment is referring to the what the patch is fixing and not to what the code is doing. Change it to something like:

```
// Ensure PtgExp carries a 4-byte operand  for the row/column of
// the array or shared formula it points at.
```

tests/test.rs
```diff
@@ -3245,6 +3245,21 @@ fn biff5_formula_ptg_ref_643() {
     assert_eq!(formulas.get_value((3, 1)), Some(&"$A$1".to_string()));
 }
 
+#[test]
+fn ptgexp_truncated_operand() {
+    // A PtgExp token (array/shared formula) whose 4-byte operand is truncated.
+    // The parser used to slice past the end of the token stream and panic with
+    // "range start index 4 out of range for slice of length 2", which took down
+    // `open_workbook` itself — the file could not be opened at all, let alone
```
jmcnamara: Again this comment is explaining what the fix is and not what is being tested. Replace it it something like:

```
// Test for xls file with a truncated PtgExp token (array/shared formula) with 2 bytes instead of 4.
```

Also call the test function `test_xls_truncated_ptgexp()`

jmcnamara (changes requested):

jmcnamara:
Thanks for the fix.

I have made some comments inline. Please address them and rebase and force push the changes to your branch. Then mark the comments as "resolved".

Also, reduce the commit message down to 1-2 lines max.

Butch78:
Added all your suggestions, thanks for everything you do with Calamine! :)

jmcnamara:
Merged. Thanks.

PR #708, opened 2026-08-11
xlsx: don't let x15:workbookPr reset the date1904 flag

### LINKED ISSUE #706: xlsx: x15:workbookPr inside extLst resets the date1904 flag (1904-system workbooks read on the 1900 epoch)
### Summary

In the xlsx reader, the `date1904` flag read correctly from `<workbookPr date1904="1"/>` is later **reset to `false`** by the `<x15:workbookPr/>` extension element that modern Excel (2013+, Windows and Mac) writes inside `<extLst>`. As a result, `has_1904_epoch()` returns `false` for real-world 1904-system workbooks and every datetime cell converts against the 1900 epoch — dates come back **four years and one day early** (−1462 days).

Tested on calamine 0.36.1.

### Mechanism

`read_workbook` matches the element by local name only:

```rust
// src/xlsx/mod.rs:517
Ok(Event::Start(e)) if e.local_name().as_ref() == b"workbookPr" => {
    self.is_1904 = match e.raw_attr(b"date1904")? {
        Some(v) => v == b"1" || v == b"true",
        None => false,
    };
}
```

`e.local_name()` strips the prefix, so `<x15:workbookPr>` (namespace `http://schemas.microsoft.com/office/spreadsheetml/2010/11/main`) also matches. That element never carries `date1904`, so the `None => false` arm overwrites the value read from the real `<workbookPr/>` earlier in the stream.

### Reproduction

Any 1904-system workbook saved by a recent Excel triggers it. Trimmed `xl/workbook.xml` from a real Excel-for-Mac file:

```xml
<workbook xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" ...>
  <fileVersion appName="xl" lastEdited="7" lowestEdited="7"/>
  <workbookPr date1904="1"/>
  <sheets><sheet name="S" sheetId="1" r:id="rId1"/></sheets>
  <extLst>
    <ext uri="{140A7094-0E35-4892-8432-C4D2E57EDEB5}" xmlns:x15="http://schemas.microsoft.com/office/spreadsheetml/2010/11/main">
      <x15:workbookPr chartTrackingRefBase="1"/>
    </ext>
  </extLst>
</workbook>
```

```rust
let mut wb: Xlsx<_> = open_workbook("date_1904_with_extlst.xlsx")?;
assert!(wb.has_1904_epoch()); // fails: returns false
// worksheet_range(...) then yields e.g. 2021-02-25 for a cell whose real value is 2025-02-26
```

Remove the `<extLst>` block and the flag is read correctly — which is presumably why the existing `tests/date_1904.xlsx` fixture doesn't catch this.

### Suggested fix

The smallest change is to stop resetting on absence — only assign when the attribute is present:

```rust
Ok(Event::Start(e)) if e.local_name().as_ref() == b"workbookPr" => {
    if let Some(v) = e.raw_attr(b"date1904")? {
        self.is_1904 = v == b"1" || v == b"true";
    }
}
```

(`<x15:workbookPr>` never carries `date1904`, so this is sufficient in practice. A stricter alternative is to only honor the `workbookPr` that is a direct child of `<workbook>` in the main spreadsheetml namespace.)

The xls (BIFF `DateMode`) and xlsb (`BrtWbProp`) readers are not affected — their flags come from binary records.

youdie006:
### Summary

Fixes #706. A genuine 1904-system xlsx workbook is read on the 1900 epoch when its `<extLst>` contains an `<x15:workbookPr>` element (which modern Excel writes), shifting every date 1462 days (4 years + 1 day) earlier.

### Root cause

`read_workbook` matches the `workbookPr` **local** name:

```rust
Ok(Event::Start(e)) if e.local_name().as_ref() == b"workbookPr" => {
    self.is_1904 = match e.raw_attr(b"date1904")? {
        Some(v) => v == b"1" || v == b"true",
        None => false,
    };
}
```

`local_name()` strips the namespace prefix, so the extension element `<x15:workbookPr>` (inside `<extLst>`) also matches. It never carries `date1904` and appears after the real `<workbookPr date1904="1"/>`, so the `None => false` arm resets `is_1904` back to `false`. `has_1904_epoch()` then returns `false` and the +1462-day offset is dropped.

### Fix

Only assign `is_1904` when the `date1904` attribute is actually present; the field already defaults to `false`, so a bare `<workbookPr/>` still means 1900-system.

### Test

Added `date_xlsx_1904_extlst` with a fixture (`tests/date_1904_extlst.xlsx`) = a 1904 workbook whose `extLst` also contains `<x15:workbookPr>`. Before the fix `has_1904_epoch()` is `false` (test fails); after, it is `true`, and the existing date tests still pass.

---

Disclosure: prepared with AI assistance (Claude); I reviewed it and verified the red-green test and the date test suite.


src/xlsx/mod.rs
```diff
@@ -515,10 +515,14 @@ impl<RS: Read + Seek> Xlsx<RS> {
                     self.sheets.push((name, path));
                 }
                 Ok(Event::Start(e)) if e.local_name().as_ref() == b"workbookPr" => {
-                    self.is_1904 = match e.raw_attr(b"date1904")? {
-                        Some(v) => v == b"1" || v == b"true",
-                        None => false,
-                    };
+                    // Only update the flag when the attribute is present. The
```
jmcnamara: This comment is referring to the what the patch is fixing and not to what the code is doing. Change it to something like 

```
// Set the 1904 date flag, if present.
```

tests/test.rs
```diff
@@ -961,6 +961,15 @@ fn date_xlsx_custom_format_with_quoted_text() {
     assert_eq!(range.get_value((3, 0)), Some(&expected));
 }
 
+#[test]
+fn date_xlsx_1904_extlst() {
+    // A 1904-system workbook whose extLst also contains an <x15:workbookPr>
+    // extension element (which shares the "workbookPr" local name but carries
+    // no date1904 attribute) must still be read on the 1904 epoch. See #706.
+    let mut xls: Xlsx<_> = wb("date_1904_extlst.xlsx");
```
jmcnamara: This doesn't need to be `mut`. Also the variable should be called `xlsx` and not `xls`.

jmcnamara (changes requested):

jmcnamara:
Thanks for the fix.

I have made some comments inline. Please address them and rebase and force push the changes to your branch. Then mark the comments as "resolved".

Also, reduce the commit message down to 1-2 lines max.

youdie006:
Pushed a fix for the red CI. The failure was mine: the test I added had a `let mut xls` that does not need `mut` (`has_1904_epoch()` takes `&self`), which `cargo clippy --all-targets --all-features -- -Dwarnings` rejects. I had copied the shape from the neighbouring `date_xlsx_1904` test, which does need `mut` because it calls `worksheet_range_at`.

One-character change, no behaviour difference. Verified locally with the same commands CI runs: `cargo clippy --all-targets --all-features -- -Dwarnings` clean, `cargo test --all-features` all green (44 + 162 + 67 passed, 0 failed).

Disclosure: I use AI assistance in my work, and I review and verify everything before it goes out.

youdie006:
Both review comments addressed — apologies, I acted on the CI failure yesterday and only picked up the `mut` half of your second note.

`src/xlsx/mod.rs`: the comment now describes what the code does rather than what the patch fixed:

```rust
// Set the 1904 date flag, if present. The attribute is
// optional here because `<x15:workbookPr>` inside `<extLst>`
// shares this local name and never carries `date1904`.
```

I kept one clause on why the attribute is optional, since without it the `if let` looks like it could just be an unconditional assignment. Happy to trim it to the single line you suggested if you would rather.

`tests/test.rs`: renamed `xls` to `xlsx`, and the `mut` is gone.

Verified with the same commands CI runs: `cargo clippy --all-targets --all-features -- -Dwarnings` clean, `cargo test --all-features` green (44 + 162 + 67 passed, 0 failed).

Disclosure: I use AI assistance in my work, and I review and verify everything before it goes out.

jmcnamara:
Merged. Thanks.

PR #710, opened 2026-08-25
feat(xlsx): add Xlsx::document_properties to read core.xml metadata

### LINKED ISSUE #149: getting the name of the author
Hello I was wondering if there was a way to find the name of the author of a sheet in the same way the file command:
file.xls: Composite Document File V2 Document, Little Endian, Os: Windows, Version 6.1, Code page: 1252, Author: AUTHOR_NAME, Last Saved By: AUTHOR_NAME, Last Printed: Wed Apr 10 07:55:25 2019, Create Time/Date: Wed Apr 10 07:57:54 2019, Last Saved Time/Date: Wed Apr 10 07:57:54 2019, Security: 0

I don't find anything in the documentation :/

ChrisJr404:
Closes #149.

This adds a `document_properties()` method on `Xlsx` that reads `docProps/core.xml` and returns the core properties of the workbook, so you can get things like the author and the created/modified dates that the `file` command shows.

The new `DocumentProperties` struct has an optional field per core property, since files created by different applications fill in different subsets. The `created`, `modified` and `last_printed` values are returned as the raw ISO 8601 (W3CDTF) strings stored in the file. I left them as strings so this doesn't pull in a date dependency just for the metadata.

It follows the same lazy pattern as `load_tables` and the pivot table methods: the part is only read when you call the method, and a workbook without a core properties part just gives back a default with every field set to `None`.

Tests cover `table-multiple.xlsx` (creator "John"), `any_sheets.xlsx` (a non-ASCII creator, to check decoding) and `date.xlsx` (which has an empty `dc:creator`, so it comes back as `None`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,8 +114,8 @@ pub use crate::ods::{Ods, OdsError};
 pub use crate::xls::{Xls, XlsError, XlsOptions};
 pub use crate::xlsb::{Xlsb, XlsbError};
 pub use crate::xlsx::{
-    expand_shared_formula, expand_shared_formula_into, Hyperlink, Xlsx, XlsxCellFormula,
-    XlsxCellFormulaMetadataRecord, XlsxCellReader, XlsxError, XlsxFormulaMetadata,
+    expand_shared_formula, expand_shared_formula_into, DocumentProperties, Hyperlink, Xlsx,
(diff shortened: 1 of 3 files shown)


jmcnamara:
This seems to be a duplicate of #685. 

If there is anything that this patch adds that the other doesn't you can add your review to #685. 

Closing.

PR #711, opened 2026-08-30
update quick-xml to 0.42.0

WindSoilder:
Hi, while I'm looking to nushell pr to update quick-xml to 0.42.0, I found out some dependencies still uses 0.41.0. https://github.com/nushell/nushell/pull/18908

Since quick-xml 0.42.0 contains some breaking changes, this pr tried to update it and fixed the breaking change, which mainly reflects changes on https://github.com/tafia/quick-xml/pull/963, which changes from `&[u8]` to `&str`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ byteorder = "1.5"
 encoding_rs = "0.8"
 fast-float2 = "0.2"
 zip = { version = "8.6", default-features = false, features = ["deflate"] }
-quick-xml = { version = "0.42.0", features = ["encoding"] }
+quick-xml = { version = "0.42", features = ["encoding"] }
 
 # Optional dependencies.
 chrono = { version = "0.4", features = ["serde"], optional = true, default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ byteorder = "1.5"
 encoding_rs = "0.8"
 fast-float2 = "0.2"
 zip = { version = "8.6", default-features = false, features = ["deflate"] }
-quick-xml = { version = "0.41", features = ["encoding"] }
+quick-xml = { version = "0.42.0", features = ["encoding"] }
 
 # Optional dependencies.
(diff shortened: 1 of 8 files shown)


jmcnamara:
Thanks. That is good work. Some comments inline and below.

Cargo.toml
```diff
@@ -22,7 +22,7 @@ byteorder = "1.5"
 encoding_rs = "0.8"
 fast-float2 = "0.2"
 zip = { version = "8.6", default-features = false, features = ["deflate"] }
-quick-xml = { version = "0.41", features = ["encoding"] }
+quick-xml = { version = "0.42.0", features = ["encoding"] }
```
jmcnamara: Leave this a `0.42` to allow for patch revisions.

src/ods.rs
```diff
@@ -672,33 +669,31 @@ where
     for attr in e.iter_raw_attrs() {
         let (key, value) = attr?;
         match key {
-            b"office:value" if !is_value_set => {
-                let v = reader.decoder().decode(value)?;
+            "office:value" if !is_value_set => {
+                let v = value;
```
jmcnamara: After your refactoring the shadowing of `v = value` isn't necessary. Just use value.

jmcnamara (changes requested):

jmcnamara:
After these changes there are a number of places where parameter types `Vec<u8>` can become `String` and `var.as_bytes().to_vec()` can become `id.to_string()` and `&[u8]` to `&str`.

See the attached patch that can be applied on top of yours, for guidance.

[711b.patch](https://github.com/user-attachments/files/31619783/711b.patch)

Cargo.toml
WindSoilder: Good idea!

src/ods.rs
WindSoilder: Got it

jmcnamara:
Merged. Thanks. That's good work.

